PR #2, opened 2019-01-02
Extend library to identify the csv, crd, and packages yamls.

awgreene:
The library needs to be able to identify csvs, crds and packages. The identify,py file
and test files were added to bring these features to the library.


kevinrizza:
@awgreene Why delete the readme?

command/tests/test_identify.py
```diff
@@ -0,0 +1,28 @@
+from unittest import TestCase
+from nose.tools import nottest
+import command.identify as id
+
+class TestGetOperatorArtifactType(TestCase):
+    @nottest
+    def id_test(self, file, expected ):
+        yaml = open(file).read()
```
MartinBasti: I recommend context manager to have file properly closed:
```python
with open(fname) as f:
    data = f.read()
```

`file` is also built-in and should't be used

command/tests/test_identify.py
```diff
@@ -0,0 +1,28 @@
...
+        actual = id.getOperatorArtifactType(yaml)
```
MartinBasti: `id` is python's built-in, it shouldn't be redefined

command/tests/test_identify.py
```diff
@@ -0,0 +1,28 @@
+from unittest import TestCase
+from nose.tools import nottest
+import command.identify as id
+
+class TestGetOperatorArtifactType(TestCase):
```
MartinBasti: If I can, I suggest to use pytest instead of nose, this whole test can be simplified in pytest using parametrize fixture

```python
@pytest.mark.parametrize('fname,expected', [
    ("command/tests/test_files/csv.yaml", "clusterServiceVersions"),
    ("command/tests/test_files/crd.yaml", "customResourceDefinitions"),
   ....
])
def test_get_operator_artifact_type(fname, expected):
    with open(fname) as f:
        yaml = f.read()
   assert getOperatorArtifactType(yaml) == expected
```

command/identify.py
```diff
@@ -0,0 +1,13 @@
+import yaml
+
+def getOperatorArtifactType(operatorArtifactString):
```
MartinBasti: This is not a pythonic name, please see PEP8 https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

command/tests/test_identify.py
awgreene: Great point - I will address this in the next commit.

command/tests/test_identify.py
awgreene: No longer using the built-in `file` variable.

command/tests/test_identify.py
awgreene: No longer using the built-in `id` variable

command/identify.py
```diff
@@ -0,0 +1,13 @@
...
+    operatorArtifact = yaml.load(operatorArtifactString)
+    if type(operatorArtifact) is dict:
```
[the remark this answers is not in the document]
awgreene: Addressed in the next commit.

command/identify.py
awgreene: Addressed in the next commit.

MartinBasti:
Thank you, LGTM

kevinrizza:
lgtm

PR #3, opened 2019-01-23
Add Real Push Method to PushCmd class

kevinrizza:
-Update push() to take defining arguments
-Create bundle from tarball
-Remove push() call from default cmd method
-Create pipfile to manage dependencies


awgreene:
@kevinrizza please add document to describe developer process given that this change introduces pipenv.

kevinrizza:
@awgreene I added a contribution file in docs/ folder that describes use of pipfile as well as relevant links.

command/push.py
```diff
@@ -1,8 +1,35 @@
...
     def __init__(self):
         pass
         
-    def push(self):
-        print("Hello, world. The push feature is currently under construction.")
+    def push(self, bundle_dir, namespace, repository, release, auth_token):
+        base64_bundle = self.__create_base64_bundle(bundle_dir, repository)
+        self.__push_to_registry(namespace, repository, release, base64_bundle, auth_token)
+
+    def __create_base64_bundle(self, bundle_dir, repository):
```
awgreene: This method accepts a bundle directory. Marketplace expects a single bundle to exist within the tar pulled from Quay.

awgreene: There is a planned usecase to use this library without providing files. We should:
1. rewrite this method to take a `String`.
2. Add a method that takes a file and provides it's content to the method mentioned in step 1.

OR

Create an issue to track the string usecase.

awgreene: Understood. LGTM. :+1:

awgreene: Understood. LGTM. :+1:

command/push.py
```diff
@@ -1,8 +1,35 @@
...
+    def __push_to_registry(self, namespace, repository, release, bundle, auth_token):
+        push_uri = 'https://quay.io/cnr/api/v1/packages/%s/%s' % (namespace, repository)
+        print("Pushing bundle to %s" % push_uri)
+        headers = {'Content-Type': 'application/json', 'Authorization': auth_token}
+        json = {'blob': bundle, 'release': release, "media_type": "helm"}
+        r = requests.post(push_uri, json=json, headers=headers)
+        if r.status_code == 200:
+            print("Pushed bundle to registry.")
+        else:
+            print("Failed to push to app registry.")
```
ralphbean: This should raise an exception, not print to stdout.

command/push.py
```diff
@@ -1,8 +1,35 @@
...
+        with tarfile.open(tarfile_name, "w:gz") as tar:
+            tar.add(bundle_dir, arcname=os.path.basename(bundle_dir))
+        with open(tarfile_name, "rb") as tarball:
+            result = tarball.read()
+        result64 = base64.b64encode(result).decode("utf-8")
+        return result64
+
+    def __push_to_registry(self, namespace, repository, release, bundle, auth_token):
+        push_uri = 'https://quay.io/cnr/api/v1/packages/%s/%s' % (namespace, repository)
+        print("Pushing bundle to %s" % push_uri)
```
ralphbean: Use the `logging` module to use different log levels?

command/push.py
kevinrizza: Good point. Updated the pr to raise an exception there.

command/push.py
kevinrizza: IMO adding a logging package like that is a change outside the scope of this pr. It's a good idea, but we should address it in a separate pr and for the entire project. For now, I just removed the info/debug print statements. We can add them in at that level when we introduce more robust logging.

kevinrizza:
@ralphbean I addressed all of your comments in the latest amended commit. Please review again. thanks!

command/push.py
```diff
@@ -1,8 +1,32 @@
...
+    def __create_base64_bundle(self, bundle_dir, repository):
+        tarfile_name = "%s.tar.gz" % repository
+        with tarfile.open(tarfile_name, "w:gz") as tar:
+            tar.add(bundle_dir, arcname=os.path.basename(bundle_dir))
+        with open(tarfile_name, "rb") as tarball:
+            result = tarball.read()
+        result64 = base64.b64encode(result).decode("utf-8")
+        return result64
+
+    def __push_to_registry(self, namespace, repository, release, bundle, auth_token):
```
MartinBasti: could you please use only one underscore, working with "private" methods in python is quite pain.

command/push.py
```diff
@@ -1,8 +1,32 @@
...
+
 class PushCmd():
     name = 'push'
 
     def __init__(self):
         pass
         
-    def push(self):
-        print("Hello, world. The push feature is currently under construction.")
+    def push(self, bundle_dir, namespace, repository, release, auth_token):
```
MartinBasti: This method is public and should have docstring

kevinrizza: The intent is that this class (and so far all classes and methods implemented outside the cli.py file) should be considered "implementation details". I haven't really written python before, but based on what I am reading these classes should be marked with a single underscore? I can mark those that way, but it seems relatively obtuse from my perspective.

I'm currently working on a separate pull request that will implement an API that is intended to be called directly and will be well documented. That api will encapsulate the implementation details of all of these methods.

command/push.py
```diff
@@ -1,8 +1,35 @@
...
+        base64_bundle = self.__create_base64_bundle(bundle_dir, repository)
+        self.__push_to_registry(namespace, repository, release, base64_bundle, auth_token)
+
+    def __create_base64_bundle(self, bundle_dir, repository):
+        tarfile_name = "%s.tar.gz" % repository
+        with tarfile.open(tarfile_name, "w:gz") as tar:
```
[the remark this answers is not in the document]
kevinrizza: Yeah, this seems very useful. I'll take a look at it.

kevinrizza:
@MartinBasti regarding your other comments:

Adding a logging package is outside the scope of this pr. We will address it in a separate pull request. I would prefer to keep our history as clean as possible.

Latest amended commit included a generic exception. Once we have a better idea for implementing custom exceptions and where they would live in the source, we will also address it. That will also be a separate pull request.

To test this specific module it will require some kind of mocking framework -- the only real method in this class will by definition make an external call so either this becomes an integration test of some kind or the endpoint needs to be mocked somehow. IMO that is also an implementation detail that is outside the scope of this pr. It is definitely necessary and will be addressed in the future, in a separate pr.

command/push.py
MartinBasti: In python there is really nothing private. Yeah Underscores means "private"

MartinBasti:
@kevinrizza 

* logging: no need for extra package, it is part of python's stdlib, it is super easy
in each module
```python
import logging

logger.info('%s successfully pushed to %s', 'something', 'something else')
```

in main function
```
logging.basicConfig(level=logging.INFO)
```


* There is nice mocking module for `requests` https://pypi.org/project/requests-mock/

command/push.py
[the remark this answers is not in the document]
kevinrizza: I took a look at this, and it seems like this will be useful, but it should be called ahead of this method to create the temporary directory that this method just accesses. (I.e. read the temporary bundle created by the build method, write it to a temporary directory. Then write the tarball to the same temp directory.) It will be addressed in a separate commit.

kevinrizza:
@MartinBasti Addressed comments in latest commit. I included the logging package since it is part of the standard library. For now, I am going to leave mocking to a future task.

PR #5, opened 2019-01-24
Generate and export API

kevinrizza:
Added courier_api.py to stub out api methods
Renamed package from command to operatorcourier for readability
on import
Updated cli.py to call api methods
Used argparse to document new cli tool methods that call api
Updated readme documentation with new uses
Added test case file for api


operatorcourier/cli.py
```diff
@@ -0,0 +1,63 @@
...
+    verify      Create a bundle and test it for correctness.
+    push        Create a bundle, test it, and push it to an app registry.
+''')
+        parser.add_argument('command', help='Subcommand to run')
+
+        args = parser.parse_args(sys.argv[1:2])
+        if not hasattr(self, args.command):
+            print('Unrecognized command')
+            parser.print_help()
+            exit(1)
```
MartinBasti: Please do `parser.error('error text')`

operatorcourier/cli.py
```diff
@@ -0,0 +1,63 @@
...
+            usage='''operator-courier <command> [<args>]
+
+These are the commands you can use:
+    verify      Create a bundle and test it for correctness.
+    push        Create a bundle, test it, and push it to an app registry.
+''')
+        parser.add_argument('command', help='Subcommand to run')
+
+        args = parser.parse_args(sys.argv[1:2])
+        if not hasattr(self, args.command):
```
MartinBasti: This can be done with subparsers https://docs.python.org/2/library/argparse.html#sub-commands

operatorcourier/cli.py
```diff
@@ -0,0 +1,63 @@
...
+        parser = argparse.ArgumentParser(description='Build, verify and push an operator bundle into external app registry.')
+        parser.add_argument('--source_dir', help='Path of your directory of yaml files to bundle. Either set this or use the files argument for bundle data.')
+        parser.add_argument('--files', required=False, help='Comma separated list of yaml files to bundle. Either set this or use the path argument for bundle data.')
+        parser.add_argument('namespace', help='Name of the Quay namespace to push operator to.')
+        parser.add_argument('repository', help='Application repository name the application is bundled for.')
+        parser.add_argument('release', help='The release version of the bundle.')
+
+        args, leftovers = parser.parse_known_args(sys.argv[2:])
+        if args.files is None and args.source_dir is None:
+            print("Neither source_dir nor files is set. One (and exactly one) of those is required to create a bundle from.")
```
MartinBasti: `parser.error('...')`

operatorcourier/courier_api.py
```diff
@@ -0,0 +1,36 @@
+"""
+operatorcourier.courier_api module
+
+This module implements the api that should be imported when using the operator-courier package
+"""
+
+from operatorcourier.build import BuildCmd
+from operatorcourier.validate import ValidateCmd
+
+def build_and_verify(source_dir="", files=""):
```
MartinBasti: Is better to use `None` as default vaules instead of empty string

operatorcourier/cli.py
```diff
@@ -0,0 +1,63 @@
+from operatorcourier import courier_api
+import argparse
+import sys
+
+def main():
+    """Generate the CLI bits
+    """
+    _CliParser()
```
MartinBasti: This looks weird to me. Object init should not do any exec operations, it should just create object.

```python
parser = _CliParser()  # just prepares parser
args = parser.parse()  # process args, can do some post parse checks

if args.command == 'push':
    do_push(args.x, args.y)
elif args.command == 'build':
   do_build(args.x, args.z)
else:
  raise RuntimeError('Unexpected commnad')
```

operatorcourier/cli.py
```diff
@@ -0,0 +1,63 @@
+from operatorcourier import courier_api
```
MartinBasti: Python module import order should be

```
stdlib imports

3rd party libs import

local app imports
```

```python
import argparse
import sys

from operatorcourier import courier_api
```

operatorcourier/courier_api.py
```diff
@@ -0,0 +1,36 @@
+"""
+operatorcourier.courier_api module
```
MartinBasti: IMHO `operatorcourier.api` is enough, is is clear form import path that API belongs to operator courier

setup.py
```diff
@@ -1,14 +1,14 @@
 from setuptools import setup
 setup(
   name='operator-courier',
-  packages=['command'],
+  packages=['operatorcourier'],
```
MartinBasti: :heart: I wanted to propose name change, but you were faster. It sounds much better.

operatorcourier/cli.py
kevinrizza: I did read about those, but I thought the pattern presented here was very readable and didn't seem to come with any specific disadvantages. What is the value of using subparsers rather than the current pattern?

MartinBasti: I prefer to use stdlib solutions

kevinrizza: Isn't this part of the stdlib? https://docs.python.org/3/library/functions.html#getattr

If there's an obvious standard I'm not following, I can modify the source to follow it. But from my reading it wasn't clear that there was a pythonic standard to add commands to a cli tool. Is there some documentation you can point me to that shows using subcommand as a standard? 

I did take a look at some implementations and it seems like using subcommand here would prevent me from using the pattern here (creating separate methods for each subcommand) which to me is very readable.

operatorcourier/cli.py
kevinrizza: Addressed in latest commit.

operatorcourier/cli.py
kevinrizza: Addressed in latest commit.

operatorcourier/courier_api.py
kevinrizza: Addressed in latest commit.

operatorcourier/cli.py
kevinrizza: Addressed in latest commit.

operatorcourier/cli.py
kevinrizza: Addressed in latest commit.

operatorcourier/courier_api.py
kevinrizza: Addressed in latest commit.

kevinrizza:
@MartinBasti Latest commit addressed comments. The pr I created earlier this morning still needs to be merged and then this pr needs to be rebased again, but it should be good to review for the requested changes.

operatorcourier/api.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if source_dir is not None: 
+        for filename in os.listdir(source_dir):
+            if filename.endswith(".yaml"):
+                with open(source_dir + "/" + filename) as f:
+                    yamls.append(f.read())
+    elif files is not None:
+        yamls = files.split(",")
+
+    bundle = BuildCmd().build_bundle(yamls)
```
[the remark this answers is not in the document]
awgreene: The CLI suggests that only the source_dir or files should be set. If the API doesn't throw an error, we may want to document that the source_dir has priority.

operatorcourier/cli.py
```diff
@@ -0,0 +1,69 @@
...
+        getattr(self, args.command)()
+
+    # Parse the verify command
+    def verify(self):
+        parser = argparse.ArgumentParser(description='Build and verify an operator bundle to test')
+        parser.add_argument('--source_dir', help='Path of your directory of yaml files to bundle. Either set this or use the files argument for bundle data.')
+        parser.add_argument('--files', help='Comma separated list of yaml files to bundle. Either set this or use the path argument for bundle data.')
+
+        args, leftovers = parser.parse_known_args(sys.argv[2:])
+        if args.files is None and args.source_dir is None:
```
awgreene: If only files or source_dir can be set, shouldn't we check if both are set? If this check isn't added, we may want to add the source_dir has greater priority

tests/test_files/valid_bundles/results/bundle.2.yaml
```diff
@@ -0,0 +1,206 @@
+data:
```
awgreene: This file seems to be identical to `tests/test_files/valid_bundles/results/bundle.1.yaml`, why does it exist?

README.md
```diff
@@ -2,6 +2,30 @@
...
+```
+from operatorcourier import api
+
+def main():
+    api.build_verify_and_push($your_namespace, $your_repository, $your_release_version, $your_quay_authtoken, source_dir="./my/folder/to/bundle/")
+```
+
+We also support passing a comma separated list of strings that make up the bundle by specifying the `files=` parameter, i.e.:
+
+`api.build_verify_and_push($your_namespace, $your_repository, $your_release_version, $your_quay_authtoken, files="MyYaml: Value1, MyYaml: Value2")`
```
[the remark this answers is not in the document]
kevinrizza: The example just above this one uses the `source_dir` parameter instead of the `files` parameter.

operatorcourier/cli.py
MartinBasti: Subparsers are standard way how to add subcommands, but let leave this as it is now

README.md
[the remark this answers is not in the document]
MartinBasti: why CSV for files instead of list?

operatorcourier/api.py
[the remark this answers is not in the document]
MartinBasti: why not to combine both inputs ?

operatorcourier/api.py
```diff
@@ -0,0 +1,56 @@
...
+    :param revision: Release version of the bundle.
+    :param source_dir: Path to local directory of yaml files to be read
+    :param files: Comma delimited string of files to create bundle with
+    """
+    
+    bundle = build_and_verify(source_dir, files)
+
+    with TemporaryDirectory() as temp_dir:
+        with open('%s/bundle.yaml' % temp_dir, 'w') as outfile:
+            yaml.dump(bundle, outfile, default_flow_style=False)
```
MartinBasti: to avoid race-conditions, you may need `outfile.flush()` after dump, to make sure that data were successfully written to file before they will be access from new filedescriptor.

operatorcourier/cli.py
MartinBasti: with parser.error you don't need exit(1) or print_help, it will exit with SystemExit=2

MartinBasti: usage text will be printed automatically with parser.error

README.md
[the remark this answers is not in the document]
kevinrizza: @MartinBasti I'm not sure I follow you? It is a list of serialized yaml files separated by commas. Do you mean the command line parameter should be serialized as a python list?

Also I assume you mean comma separated values (CSV) and not ClusterServiceVersion (CSV), we should probably be more specific in this case.

operatorcourier/api.py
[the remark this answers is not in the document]
kevinrizza: Not sure I follow. They are very different inputs. one is a single string that is a path, the other is a list of serialized yaml. do you mean they should just be one parameter and we have code that determines if it is a path or not?

operatorcourier/api.py
kevinrizza: @MartinBasti I thought the TemporaryDirectory() method creates a new, unique directory path? In that case, what race condition could we hit if the file and path are unique?

README.md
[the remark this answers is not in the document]
MartinBasti: But why do you use string with comma-separated-values? Isn't easier to use just standard list/tuple instead having complications with parsing string?

Just question, what happen when yaml will contain textual value with comma inside?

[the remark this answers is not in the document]
MartinBasti: also `files` name is tricky, when I see that, I think it expects filepath or fileobject, not content

operatorcourier/api.py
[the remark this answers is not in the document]
MartinBasti: No, I want two parameters, just append "files" to yamls via `yaml.extend()` when defined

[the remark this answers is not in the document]
kevinrizza: @MartinBasti are you planning on using the cli interface or importing the api module? Your team is the only reason we were planning on supporting passing the files themselves rather than a directory parameter, and this method is only written that way to support the cli use case. If you aren't going to use the cli, I will just make this a python list parameter and remove that param from the cli.

kevinrizza:
@awgreene @MartinBasti latest commit addressed all comments. please review again.

operatorcourier/api.py
```diff
@@ -0,0 +1,66 @@
...
+
+    :param source_dir: Path to local directory of yaml files to be read.
+    :param yamls: List of yaml strings to create bundle with
+    """
+
+    yaml_files = []
+
+    if source_dir is not None: 
+        for filename in os.listdir(source_dir):
+            if filename.endswith(".yaml"):
```
MartinBasti: is possible to have `.yml` ending? Both `.yaml` and `.yml` are used with files associated to YAML content

operatorcourier/api.py
```diff
@@ -0,0 +1,66 @@
...
+                with open(source_dir + "/" + filename) as f:
+                    yaml_files.append(f.read())
+    elif yamls is not None:
+        yaml_files = yamls
```
MartinBasti: why not `yaml_files.extend(yamls)` to respect both options?

If options are mutually exclusive, you should check and raise an exception when both options were used.

operatorcourier/api.py
```diff
@@ -0,0 +1,66 @@
...
+
+    bundle = BuildCmd().build_bundle(yaml_files)
+
+    valid = ValidateCmd().validate(bundle)
+
+    if not valid:
+        bundle = None
+        logger.error("Bundle failed validation.")
```
MartinBasti: Shouldn't be this an exception?

operatorcourier/api.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if bundle is not None:
+        with TemporaryDirectory() as temp_dir:
+            with open('%s/bundle.yaml' % temp_dir, 'w') as outfile:
+                yaml.dump(bundle, outfile, default_flow_style=False)
+                outfile.flush()
+
+            PushCmd().push(temp_dir, namespace, repository, revision, token)
+    else:
+        logger.error("Bundle is invalid. Will not attempt to push.")
```
MartinBasti: IMO this should be exception

tests/test_api.py
```diff
@@ -0,0 +1,32 @@
+from unittest import TestCase
+import pytest
+import yaml
+from operatorcourier import api
+
+@pytest.mark.parametrize('dir,expected', [
+("tests/test_files/bundles/api/bundle1", "tests/test_files/bundles/api/bundle1/results/bundle.yaml"),
+])
+def test_make_bundle(dir, expected):
+    bundle = api.build_and_verify(source_dir=dir)
```
MartinBasti: `dir` is python's built-in, please use different name for variable

operatorcourier/api.py
kevinrizza: updated to include them.

operatorcourier/api.py
kevinrizza: IMO I am already concerned about the supportability of passing the files as a list (OLM is already implementing a change to use a nested hierarchy with different crds associated with specific versions of csvs), so I would like to keep them separate. I added a check that prevents both parameters from being passed.

operatorcourier/api.py
kevinrizza: This api is informational, so passing bad files shouldn't be an exception. A log message is the correct output. The push method should test this and throw an exception when validation fails though.

operatorcourier/api.py
kevinrizza: Agreed, raising an exception here now.

tests/test_api.py
kevinrizza: updated.

kevinrizza:
@MartinBasti addressed your comments.

PR #6, opened 2019-01-29
Add real build method

kevinrizza:
-Create build_bundle method
-Update identify package to better handle errors
-Add build method tests
-Add yaml and test packages to pipfile

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,18 +1,18 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "a0e63f8a0d1e3df046dc19b3ffbaaedfa151afc12af5a5b960ae7393952f8679"
+            "sha256": "4ad510585c132fabcfbceac4231a8f427c1c334e0fe34fab1b5219277de706fe"
         },
         "host-environment-markers": {
             "implementation_name": "cpython",
-            "implementation_version": "3.6.6",
+            "implementation_version": "3.6.8",
             "os_name": "posix",
             "platform_machine": "x86_64",
             "platform_python_implementation": "CPython",
             "platform_release": "4.18.16-200.fc28.x86_64",
             "platform_system": "Linux",
             "platform_version": "#1 SMP Sat Oct 20 23:53:47 UTC 2018",
-            "python_full_version": "3.6.6",
+            "python_full_version": "3.6.8",
             "python_version": "3.6",
             "sys_platform": "linux"
         },
@@ -27,6 +27,20 @@
         ]
     },
     "default": {
+        "atomicwrites": {
+            "hashes": [
+                "sha256:0312ad34fcad8fac3704d441f7b317e50af620823353ec657a53e981f92920c0",
+                "sha256:ec9ae8adaae229e4f8446952d204a3e4b5fdd2d099f9be3aaf556120135fb3ee"
+            ],
+            "version": "==1.2.1"
+        },
+        "attrs": {
+            "hashes": [
+                "sha256:ca4be454458f9dec299268d472aaa5a11f67a4ff70093396e1ceae9c76cf4bbb",
+                "sha256:10cbf6e27dbce8c30807caf056c8eb50917e0eaafe86347671b57254006c3e69"
+            ],
+            "version": "==18.2.0"
+        },
         "certifi": {
             "hashes": [
                 "sha256:993f830721089fef441cdfeb4b2c8c9df86f0c63239f06bd025a76a7daddb033",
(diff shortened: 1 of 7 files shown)


command/build.py
```diff
@@ -1,8 +1,39 @@
+import yaml as pyYaml
+import command.identify as identify
+
 class BuildCmd():
     name = 'build'
-
-    def __init__(self):	
+
```
SamiSousa: Remove unnecessary whitespace

command/build.py
```diff
@@ -1,8 +1,39 @@
...
+            )
+        )
+        return operatorArtifactsTemplate
     
-    def build(self):
-        print("Hello, world. The build feature is currently under construction.")
+    def _get_field_entry(self, yamlContent):
+        yaml_type = identify.get_operator_artifact_type(yamlContent)
+        if yaml_type == "ClusterServiceVersion" or yaml_type == "CustomResourceDefinition" or yaml_type == "Package":
+            return yaml_type[0:1].lower() + yaml_type[1:] + 's'
```
SamiSousa: Do we need the if statement? The way this is used in `_updateArtifact` requires an output, so in the case that `yaml_type` is an unexpected value the program errors out; is this a "pythonic" way of erroring out? Should there be an error log as well?

awgreene: Actually the identify.get_operator_artifact_type method will raise an exception if the file isn't identified as a CSV, CRD, or Pacakge YAML - your points that the `if` statement isn't needed and logging are still valid.

command/build.py
```diff
@@ -1,8 +1,39 @@
...
+
+    def _updateArtifact(self, operatorArtifact, yamlContent):
```
awgreene: I assume the underscore means this is a private method, in these cases do we not add method descriptions?

awgreene: can we change operatorArtifact to Bundle? Or operatorBundle?

awgreene: Can we change the name of this method from _updateArtifact to  _updateBundle?

command/build.py
kevinrizza: @SamiSousa What is unnecessary about this whitespace? It makes the pace between setting name and the init method more readable.

command/build.py
kevinrizza: @awgreene yes that is the case.

command/build.py
SamiSousa: The newline is fine, but it's followed by 4 spaces which are unnecessary.

command/build.py
kevinrizza: I don't think it makes sense to add a log when the raised exception will give the same information. I did address the comment related to removing the if statement --- we will never hit it.

kevinrizza:
Closed by accident.....

command/build.py
kevinrizza: Addressed in latest commit.

command/build.py
kevinrizza: Addressed in latest commit.

command/build.py
kevinrizza: Addressed in latest commit.

kevinrizza: Addressed in latest commit.

command/build.py
```diff
@@ -1,8 +1,38 @@
...
+
 class BuildCmd():
     name = 'build'
 
-    def __init__(self):	
+    def __init__(self):
         pass
+
+    def _get_empty_bundle(self):
+        operatorBundleTemplate = dict(
```
awgreene: Could we simply return the dict?

command/identify.py
```diff
@@ -1,13 +1,15 @@
 import yaml
 
 def get_operator_artifact_type(operatorArtifactString):
+    """get_operator_artifact_type takes a yaml string and determines if it is one of the expected bundle types
```
awgreene: Could you change this to mention that it checks for the expected CSV, CRD, or Package type?

command/build.py
kevinrizza: True. Updated in latest commit.

command/identify.py
kevinrizza: Updated in latest commit.

command/build.py
```diff
@@ -1,8 +1,37 @@
...
+                customResourceDefinitions = [],
+                clusterServiceVersions = [],
+                packages = [],
+            )
+        )
     
-    def build(self):
-        print("Hello, world. The build feature is currently under construction.")
+    def _get_field_entry(self, yamlContent):
+        yaml_type = identify.get_operator_artifact_type(yamlContent)
```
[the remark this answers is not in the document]
aravindhp: Given we are going to be using `operator-registry`, we should be adopting [definitions](https://github.com/operator-framework/operator-registry#manifest-format) that are being used there.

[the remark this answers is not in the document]
kevinrizza: @SamiSousa Artifact in this context (identify.py) does not refer to the bundle, it refers to any subsection of yaml that is being created. Since we don't have a term for a generic 'component' I think `Artifact` is fine in this context.

command/build.py
```diff
@@ -1,8 +1,37 @@
...
+        return yaml_type[0:1].lower() + yaml_type[1:] + 's'
+
+    def _updateBundle(self, operatorBundle, yamlContent):
+        operatorBundle["data"][self._get_field_entry(yamlContent)].append(pyYaml.load(yamlContent))
```
MartinBasti: Please always use `pyYaml.safe_load` to avoid security issues, unless you really need `yaml.load` and you can guarantee that `yamlContent` is safe.

command/build.py
```diff
@@ -1,8 +1,37 @@
+import yaml as pyYaml
+import command.identify as identify
+
 class BuildCmd():
     name = 'build'
 
-    def __init__(self):	
+    def __init__(self):
```
MartinBasti: empty `__init__` can be omitted

command/identify.py
```diff
@@ -1,13 +1,16 @@
...
         if "packageName" in operatorArtifact:
-            return "packages"
+            return "Package"
         elif "kind" in operatorArtifact:
-            if operatorArtifact["kind"] == "ClusterServiceVersion": 
-                return "clusterServiceVersions"
-            elif operatorArtifact["kind"] == "CustomResourceDefinition":
-                return "customResourceDefinitions"
-    return "invalid"
+            if operatorArtifact["kind"] == "ClusterServiceVersion" or operatorArtifact["kind"] == "CustomResourceDefinition":
```
MartinBasti: just protip `if operatorArtifact["kind"] in ("ClusterServiceVersion", "CustomResourceDefinition"):`

tests/test_build.py
```diff
@@ -0,0 +1,16 @@
+from unittest import TestCase
+import pytest
+from command.build import BuildCmd
+
+
+def test_create_bundle():
+    paths = ["tests/test_files/csv.yaml", "tests/test_files/crd.yaml", "tests/test_files/package.yaml"]
+    yamls = []
+    for x in range(3):
```
MartinBasti:
```python
for path in paths:
    with open(path) as f:
        ...
```

command/build.py
```diff
@@ -1,8 +1,37 @@
+import yaml as pyYaml
```
MartinBasti: in previous PR you used just `yaml` this is quite inconsistent naming

Pipfile.lock
```diff
@@ -48,13 +62,65 @@
...
+            "version": "==1.7.0"
+        },
+        "pytest": {
+            "hashes": [
+                "sha256:41568ea7ecb4a68d7f63837cf65b92ce8d0105e43196ff2b26622995bb3dc4b2",
+                "sha256:c3c573a29d7c9547fb90217ece8a8843aa0c1328a797e200290dc3d0b4b823be"
+            ],
+            "version": "==4.1.1"
+        },
+        "pyyaml": {
```
MartinBasti: Please use this yaml module, it is production stable, your version is in beta according pypi
https://pypi.org/project/PyYAML/

kevinrizza: @MartinBasti What do you mean? The linked version (3.13) is listed as stable and is the version we are importing and using.

kevinrizza:
@MartinBasti Addressed your comments in latest commit.

command/build.py
kevinrizza: @MartinBasti addressed in latest commit.

tests/test_build.py
kevinrizza: @MartinBasti addressed in latest commit.

command/identify.py
kevinrizza: @MartinBasti addressed in latest commit.

command/build.py
kevinrizza: @MartinBasti addressed in latest commit.

command/build.py
kevinrizza: @MartinBasti addressed in latest commit.

Pipfile.lock
MartinBasti: Ah sorry, it was unexpected to see it in lowercase, I thought that this was https://pypi.org/project/pyaml/

So it is ok, sorry for the noise

command/build.py
```diff
@@ -1,8 +1,32 @@
-class BuildCmd():
-    name = 'build'
+import yaml
+import command.identify as identify
```
MartinBasti: how about `from command import identify` ?

MartinBasti:
LGTM

command/build.py
```diff
@@ -1,8 +1,32 @@
...
+        )
     
-    def build(self):
-        print("Hello, world. The build feature is currently under construction.")
+    def _get_field_entry(self, yamlContent):
+        yaml_type = identify.get_operator_artifact_type(yamlContent)
+        return yaml_type[0:1].lower() + yaml_type[1:] + 's'
+
+    def _updateBundle(self, operatorBundle, yamlContent):
+        operatorBundle["data"][self._get_field_entry(yamlContent)].append(yaml.safe_load(yamlContent))
```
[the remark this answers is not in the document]
MartinBasti: Please see example: https://security.openstack.org/guidelines/dg_avoid-dangerous-input-parsing-libraries.html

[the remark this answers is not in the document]
awgreene: Thanks @MartinBasti !

command/build.py
kevinrizza: addressed in latest commit

PR #7, opened 2019-01-30
Add validate method

kevinrizza:
-Update validate.py with real validation logic
-Add test file to verify bundles

### DIFF
--- a/command/validate.py
+++ b/command/validate.py
@@ -1,8 +1,239 @@
+import logging
+import json
+
+logger = logging.getLogger(__name__)
+
 class ValidateCmd():
-    name = 'validate'
+    dataKey = "data"
+    crdKey = "customResourceDefinitions"
+    csvKey = "clusterServiceVersions"
+    pkgsKey = "packages"
 
     def __init__(self):
         pass
         
-    def validate(self):
-        print("Hello, world. The validate feature is currently under construction.")
+    def validate(self, bundle):
+        """validate takes a bundle as a dictionary and returns a boolean value that
+        describes if the bundle is valid. It also logs verification information when
+        there are invalid sections in the bundle.
+
+        :param bundle: Dictionary of bundle value
+        """
+        logger.info("Validating bundle.")
+
+        return self._bundle_validation(bundle)
+
+    def _bundle_validation(self, bundle):
+        validationDict = dict()
+
+        if self.dataKey not in bundle:
+            logger.error("Bundle does not contain base data field.")
+            #break here because there's nothing else to learn
+            return False
+
+        bundleData = bundle[self.dataKey]
+
+        validationDict[self.crdKey] = self._type_validation(bundleData, self.crdKey, self._crd_validation, False)  
+        validationDict[self.csvKey] = self._type_validation(bundleData, self.csvKey, self._csv_validation, True)
+        validationDict[self.pkgsKey] = self._type_validation(bundleData, self.pkgsKey, self._pkgs_validation, True)
+
+        valid = True
+        for key, value in validationDict.items():
+            if value is False:
+                valid = False
+        
+        return valid
+
+    def _crd_validation(self, bundleData):
+        logger.info("Validating custom resource definitions.")
+        valid = True
+
+        crds = bundleData[self.crdKey]
+
+        for crd in crds:
+            if "metadata" in crd:
+                if "name" in crd["metadata"]:
+                    logger.info("Evaluating crd %s", crd["metadata"]["name"])
+                else:
+                    logger.error("crd metadata.name not defined.")
+                    valid = False
+            else:
+                logger.error("crd metadata not defined.")
+                valid = False
+
+            if "apiVersion" not in crd:
+                logger.error("crd apiVersion not defined.")
+                valid = False
+
+            if "spec" not in crd:
+                logger.error("crd spec not defined.")
+                valid = False
+
+        return valid
+
+    def _csv_validation(self, bundleData):
+        valid = True
+        logger.info("Validating cluster service versions.")
+
+        csvs = bundleData[self.csvKey]
+
+        for csv in csvs:
+            if "metadata" in csv:
+                if self._csv_metadata_validation(csv["metadata"]) is False:
+                    valid = False
+            else:
+                logger.error("csv metadata not defined.")
+                valid = False
+
+            if "apiVersion" not in csv:
+                logger.error("csv apiVersion not defined.")
+                valid = False
+
+            if "spec" in csv:
+                if self._csv_spec_validation(csv["spec"], bundleData) is False:
+                    valid = False
+            else:
+                logger.error("csv spec not defined.")
+                valid = False
+
+        return valid
+
+    def _csv_spec_validation(self, spec, bundleData):
+        valid = True
+
+        warnSpecList = ["displayName", "description", "icon", "version", "provider", "maturity"]
+        
+        for item in warnSpecList:
+            if item not in spec:
+                logger.warning("csv spec.%s not defined" % item)
+
+        if "installModes" not in spec:
+            logger.error("csv spec.installModes not defined" % item)
+            valid = False
+
+        if "install" not in spec:
+            logger.error("csv spec.install not defined" % item)
+            valid = False
+
+        if "customresourcedefinitions" in spec:
+            customresourcedefinitions = spec["customresourcedefinitions"]
+
+            crdList = []
+            for crd in bundleData[self.crdKey]:
+                try:
+                    name = crd["metadata"]["name"]
+                    crdList.append(name)
+                except KeyError:
+                    pass
+
+            if "owned" not in customresourcedefinitions:
+                logger.error("spec.customresourcedefinitions.owned not defined for csv")
+                valid = False
+            else:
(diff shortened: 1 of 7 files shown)


command/validate.py
```diff
@@ -1,8 +1,243 @@
...
+
+        return valid
+
+    def _type_validation(self, bundleData, typeName, validator):
+        valid = True
+        if typeName in bundleData:
+            if len(bundleData[typeName]) != 0:
+                valid = validator(bundleData)
+            else:
+                logger.error("Bundle does not contain any %s." % typeName)
```
SamiSousa: It's possible for a valid bundle to be created without any CRDs if the CSV doesn't list any. An example is the Service Catalog operator: https://github.com/operator-framework/community-operators/blob/master/community-operators/bundles/svcat/bundle.0.1.34.yaml#L7

command/validate.py
```diff
@@ -1,8 +1,243 @@
...
+            crdList = []
+            for crd in bundleData[self.crdKey]:
+                try:
+                    name = crd["metadata"]["name"]
+                    crdList.append(name)
+                except:
+                    pass
+
+            if "owned" not in customresourcedefinitions:
+                logger.error("spec.customresourcedefinitions.owned not defined for csv")
```
SamiSousa: It's possible for an operator to not use any CRDs.

command/validate.py
```diff
@@ -1,8 +1,243 @@
...
+                if item not in annotations:
+                    logger.warning("csv metadata.annotations.%s not defined" % item)
+
+            # check certified value's type in particular. should be string, not bool
+            if "certified" not in annotations:
+                logger.warning("csv metadata.annotations.certified not defined.")
+            else:
+                isString = isinstance(annotations["certified"], str)
+                if not isString:
+                    logger.error("metadata.annotations.certified is not of type string")
```
SamiSousa: Thank you for including this! :)

tests/test_files/csv.yaml
```diff
@@ -5,6 +5,16 @@ kind: ClusterServiceVersion
 metadata:
   name: marketplace-operator.v0.0.1
   namespace: placeholder
+  annotations:
+    categories: openshift required
+    certified: "true"
```
SamiSousa: nit: Noticed for the other test files, the strings in the `certified` field used single quotes rather than double quotes. These should follow a single convention.

tests/test_files/csv.yaml
```diff
@@ -5,6 +5,16 @@ kind: ClusterServiceVersion
...
+    containerImage: quay.io/openshift/origin-operator-marketplace:latest
+    createdAt: 2019/11/15
```
SamiSousa: The `createdAt` field should use the `yyyy-mm-ddThh:mm:ssZ` format as described in the marketplace operator repo: https://github.com/operator-framework/operator-marketplace/blob/master/docs/marketplace-required-csv-annotations.md

command/validate.py
```diff
@@ -1,8 +1,243 @@
...
+        bundleData = bundle[self.dataKey]
+
+        validationDict[self.crdKey] = self._type_validation(bundleData, self.crdKey, self._crd_validation)  
+        validationDict[self.csvKey] = self._type_validation(bundleData, self.csvKey, self._csv_validation)
+        validationDict[self.pkgsKey] = self._type_validation(bundleData, self.pkgsKey, self._pkgs_validation)
+
+        valid = True
+        for key, value in validationDict.items():
+            if value is False:
+                valid = False
```
MartinBasti: Maybe log which value is wrong?

MartinBasti: Ah it is logged elsewhere

command/validate.py
kevinrizza: Addressed in latest commit.

command/validate.py
kevinrizza: Addressed in latest commit.

tests/test_files/csv.yaml
kevinrizza: Addressed in latest commit.

tests/test_files/csv.yaml
kevinrizza: Addressed in latest commit.

command/validate.py
```diff
@@ -1,8 +1,243 @@
...
+
+        if "customresourcedefinitions" in spec:
+            customresourcedefinitions = spec["customresourcedefinitions"]
+
+            crdList = []
+            for crd in bundleData[self.crdKey]:
+                try:
+                    name = crd["metadata"]["name"]
+                    crdList.append(name)
+                except:
```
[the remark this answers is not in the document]
kevinrizza: Addressed in latest commit.

command/validate.py
```diff
@@ -1,8 +1,243 @@
...
+                isString = isinstance(annotations["certified"], str)
+                if not isString:
+                    logger.error("metadata.annotations.certified is not of type string")
+                    valid = False
+
+            # if alm-examples is defined, check that its value is valid json
+            if "alm-examples" in annotations:
+                try:
+                    valid_json = json.loads(annotations["alm-examples"])
+                except:
```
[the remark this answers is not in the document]
kevinrizza: Addressed in latest commit.

command/validate.py
```diff
@@ -1,8 +1,243 @@
...
+                if len(channels) == 0:
+                    logger.error("no package channels defined.")
+                    valid = False
+                else:
+                    csvNames = []
+                    for csv in bundleData[self.csvKey]:
+                        try:
+                            csvName = csv["metadata"]["name"]
+                            csvNames.append(csvName)
+                        except:
```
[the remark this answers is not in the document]
kevinrizza: Addressed in latest commit.

tests/test_validate.py
```diff
@@ -0,0 +1,17 @@
+from unittest import TestCase
```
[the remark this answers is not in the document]
kevinrizza: Addressed in latest commit.

kevinrizza:
@MartinBasti @SamiSousa Addressed all of your comments in latest commit. Please review again. Thanks!

command/validate.py
```diff
@@ -1,8 +1,239 @@
...
+        """
+        logger.info("Validating bundle.")
+
+        return self._bundle_validation(bundle)
+
+    def _bundle_validation(self, bundle):
+        validationDict = dict()
+
+        if self.dataKey not in bundle:
+            logger.error("Bundle does not contain base data field.")
```
awgreene: In L8 we refer to the CSV/CRD/Package as Keys yet here we refer to them as fields. Should we converge on a single naming convention?

command/validate.py
```diff
@@ -1,8 +1,239 @@
+import logging
+import json
+
+logger = logging.getLogger(__name__)
+
 class ValidateCmd():
-    name = 'validate'
+    dataKey = "data"
```
awgreene: These are also used in the build method, could we put these values somewhere shared?

command/validate.py
```diff
@@ -1,8 +1,239 @@
...
+                    crdList.append(name)
+                except KeyError:
+                    pass
+
+            if "owned" not in customresourcedefinitions:
+                logger.error("spec.customresourcedefinitions.owned not defined for csv")
+                valid = False
+            else:
+                for crd in customresourcedefinitions["owned"]:
+                    if "name" not in crd:
```
awgreene: Should we also check the versions of the CRD defined in the CSV and the CRDs provided in the bundle?

command/validate.py
awgreene: Best feature by far.

command/validate.py
kevinrizza: @awgreene the word `key` is used here to describe the mechanical feature of key,value in the `dict` type. Whenever the code refers to `key` it is describing a hardcoded string that is used to extract specific data. Field is used to describe the concept of a generic data field on a type, which is a different concept. They aren't interchangeable, so I am using different words here.

command/validate.py
kevinrizza: IMO limiting cross dependencies is important in these modules --- and I'm not sure I see value to creating a "shared" module just for these strings. If they shared something more fundamental though and a shared module was required we could always refactor this to move these definitions.

command/validate.py
kevinrizza: @awgreene We can, but could we make that a separate feature / effort for later? It might be useful to have someone else work on this to make changes at a later point, and I think having every use case defined right now in the validator isn't necessary.

command/validate.py
awgreene: Cross dependencies should certainly be avoided - my  hesitation stems from a potential change to the bundle structure, which could come from outside teams, where we would need to update these keys in multiple areas (in this case the build, identify, and validate classes). This isn't a blocking issue, but something we may want to add as an enhancement.

command/validate.py
awgreene: I recognize the distinction and agree that using the term `field` would be easier for users to absorb. LGTM.

command/validate.py
awgreene: I absolutely agree - this wasn't mean't to be a blocking issue. Let's add it later as an enhancement.

command/validate.py
awgreene: The suggestion of a bundle class sounds good to me.

Issue #15, opened 2019-02-14
exit code not set correctly if validation fails


scoheb:
`[root@0d92433c567b test]# operator-courier verify .`
`csv metadata.annotations not defined.`
`csv spec.icon not defined`
`[root@0d92433c567b test]# echo $?`
`0`

SamiSousa:
I believe in this case the output is warning that some recommended fields are empty. This isn't an error and verification passes since the CSV is valid.

scoheb:
That makes sense. 
It may be more helpful if each warning is prefixed with `WARNING: `

kevinrizza:
This issue is resolved by https://github.com/operator-framework/operator-courier/commit/477ecbc0eb4d00fbbf3b7c6dc5785d521e67be69

PR #19, opened 2019-02-15
Add logging level and module name to logs

awgreene:
Addresses issue #15


awgreene:
Example: `csv metadata.annotations.categories not defined` -> now prints as `WARNING:operatorcourier.validate:csv metadata.annotations.categories not defined`

scoheb:
@kevinrizza is this part of 1.0.1?

kevinrizza:
@scoheb It is not, 1.0.1 was released before this was merged.

PR #21, opened 2019-02-18
Pass error msg from quay.io to user

MartinBasti:
Don't hide reason why push failed.

### DIFF
--- a/operatorcourier/push.py
+++ b/operatorcourier/push.py
@@ -44,4 +44,13 @@ class PushCmd():
         r = requests.post(push_uri, json=json, headers=headers)
         if r.status_code != 200:
             logger.error(r.text)
-            raise ValueError("Registry api returned non 200 response. Failed to push to app registry.")
+
+            msg = 'Failed to get error details'
+            try:
+                error = r.json()['error']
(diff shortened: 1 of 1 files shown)


operatorcourier/push.py
```diff
@@ -44,4 +44,13 @@ def _push_to_registry(self, namespace, repository, release, bundle, auth_token):
         r = requests.post(push_uri, json=json, headers=headers)
         if r.status_code != 200:
             logger.error(r.text)
-            raise ValueError("Registry api returned non 200 response. Failed to push to app registry.")
+
+            msg = 'Failed to get error details'
+            try:
+                error = r.json()['error']
+                msg = "Quay.io answered: '{}' ({})".format(
+                    error['code'], error['message'])
```
twaugh: Are these the right way round?

MartinBasti: IMO yes, do you thing that `Quay.io answered: long message ('shortcode')` is better? (I have no strong preference)

twaugh: I have no strong preference either, just thought you probably wanted the code within the parentheses. :)

MartinBasti: I'd leave it as is then

PR #22, opened 2019-02-18
Update README.md with example

kevinrizza:


README.md
```diff
@@ -5,6 +5,29 @@ The Operator Courier is used to build, validate and push Operator Artifacts.
 It is currently in a pre-alpha stage.
 
 ## Usage
+
+### Command Line Interface
+To install the operator courier project to use from the command line, just install the latest release from [PyPI](https://pypi.org/project/operator-courier/):
+
+```bash
+$ pip3 install operator-courier
+$ operator-courier
```
aravindhp: This seems to imply executing `operator-courier` is required to install the package. I suggest leaving it out.

README.md
```diff
@@ -5,6 +5,29 @@ The Operator Courier is used to build, validate and push Operator Artifacts.
...
+```
+
+Once the project is installed, to generate a bundle and push it to a quay.io app registry just use `operator-courier push`
```
aravindhp: `generate a bundle` seems to imply the tool is generating the files. What is required here is a note about how the user should organize the files.

kevinrizza: @aravindhp courier does generate the bundle yaml. I can add more specifics here though to define exactly what is expected and what is generated.

aravindhp: Are we re-defining what a [bundle](https://github.com/operator-framework/operator-registry#manifest-format) is? But yes, calling out what is expected will help.

kevinrizza: @aravindhp Here we are using the word `bundle` to describe the yaml blob that is pushed up to quay (similarly to how it is used in the community-operators repo). If and when we update courier to use the nested bundles, we should update our readme and source to use the correct terms -- AFAIK there is no correct term right now for the yaml blob using the new nested format.

README.md
kevinrizza: updated.

README.md
kevinrizza: updated.

README.md
```diff
@@ -18,17 +40,9 @@ We also support passing a list of strings that make up the bundle by specifying
 
 `api.build_verify_and_push($your_namespace, $your_repository, $your_release_version, $your_quay_authtoken, yamls=$your_yamls_list)`
 
-Not only that, but we also support the use of the tool as a command line interface. Just invoke `operator-courier` from a shell:
-
-```bash
-$ operator-courier verify --source_dir="./temp/bundles/couchbase/"
```
SamiSousa: Do we want to omit mention of the verify command? I agree that the focus should be on the push command, but this entirely removes mention of it.

aravindhp: That's a good point. It makes sens to call out how to verify it too.

kevinrizza: included in latest push.

Issue #16, opened 2019-02-14
malformed yaml file causes courier to blowup


scoheb:
Purposely moved a `-` in a crd.yml file and ran verify:

> [root@0d92433c567b test]# operator-courier verify .
Traceback (most recent call last):
  File "/usr/local/bin/operator-courier", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/cli.py", line 9, in main
    parser.parse()
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/cli.py", line 35, in parse
    getattr(self, args.command)()
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/cli.py", line 43, in verify
    api.build_and_verify(source_dir=args.source_dir)
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/api.py", line 39, in build_and_verify
    bundle = BuildCmd().build_bundle(yaml_files)
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/build.py", line 30, in build_bundle
    bundle = self._updateBundle(bundle, item)
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/build.py", line 19, in _updateBundle
    operatorBundle["data"][self._get_field_entry(yamlContent)].append(yaml.safe_load(yamlContent))
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/build.py", line 15, in _get_field_entry
    yaml_type = identify.get_operator_artifact_type(yamlContent)
  File "/usr/local/lib/python3.6/site-packages/operatorcourier/identify.py", line 9, in get_operator_artifact_type
    operatorArtifact = yaml.safe_load(operatorArtifactString)
  File "/usr/local/lib64/python3.6/site-packages/yaml/__init__.py", line 94, in safe_load
    return load(stream, SafeLoader)
  File "/usr/local/lib64/python3.6/site-packages/yaml/__init__.py", line 72, in load
    return loader.get_single_data()
  File "/usr/local/lib64/python3.6/site-packages/yaml/constructor.py", line 35, in get_single_data
    node = self.get_single_node()
  File "/usr/local/lib64/python3.6/site-packages/yaml/composer.py", line 36, in get_single_node
    document = self.compose_document()
  File "/usr/local/lib64/python3.6/site-packages/yaml/composer.py", line 55, in compose_document
    node = self.compose_node(None, None)
[...]

PR #27, opened 2019-02-21
Add option to print validation errors and warnings to file

awgreene:

PR #35, opened 2019-03-01
Handle malformed yaml files

jeremylinlin:
- This PR fixes issue https://github.com/operator-framework/operator-courier/issues/16
- Log with error when a malformed YAML is being parsed
- Log with info when valid YAML has been parsed
- Suppress error traceback when calling the CLI


operatorcourier/build.py
```diff
@@ -3,14 +3,14 @@
 
 class BuildCmd():
     def _get_empty_bundle(self):
-         return dict(
+        return dict(
```
kevinrizza: There are no actual changes in this file. Please remove this formatting only change as it just adds noise to the commit log.

operatorcourier/identify.py
```diff
@@ -1,16 +1,27 @@
...
-        if "packageName" in operatorArtifact:
-            return "Package"
-        elif "kind" in operatorArtifact:
-            if operatorArtifact["kind"] in ("ClusterServiceVersion", "CustomResourceDefinition"):
-                return operatorArtifact["kind"]
-    raise ValueError('Courier requires valid CSV, CRD, and Package files')
+    try:
+        operatorArtifact = safe_load(operatorArtifactString)
+    except MarkedYAMLError:
+        logger.error("Invalid YAML file")
```
kevinrizza: Can we mention which file couldn't be parsed here?

operatorcourier/identify.py
```diff
@@ -1,16 +1,27 @@
...
+        print('Courier requires valid input YAML files')
```
kevinrizza: Please avoid using print() and instead only use the logger methods.

operatorcourier/identify.py
```diff
@@ -1,16 +1,27 @@
...
+        exit(0)
```
kevinrizza: Rather than `exit(0)` here, can we raise an exception like later in the file?

operatorcourier/validate.py
```diff
@@ -33,16 +33,11 @@ def _bundle_validation(self, bundle):
...
         validationDict[self.csvKey] = self._type_validation(bundleData, self.csvKey, self._csv_validation, True)
         validationDict[self.pkgsKey] = self._type_validation(bundleData, self.pkgsKey, self._pkgs_validation, True)
 
-        valid = True
-        for key, value in validationDict.items():
-            if value is False:
-                valid = False
-        
-        return valid
+        return False not in validationDict.values()
```
kevinrizza: How is this change related to the malformed yaml? Please limit this pr to a change related to the comments mentioned in the pr and commit message.

tests/test_api.py
```diff
@@ -14,11 +14,14 @@ def test_make_bundle(directory, expected):
...
-(["tests/test_files/bundles/api/bundle1/crd.yml",
-"tests/test_files/bundles/api/bundle1/csv.yaml",
-"tests/test_files/bundles/api/bundle1/packages.yaml"],
-"tests/test_files/bundles/api/bundle1/results/bundle.yaml"),
+    ([
+         "tests/test_files/bundles/api/bundle1/crd.yml",
+         "tests/test_files/bundles/api/bundle1/csv.yaml",
+         "tests/test_files/bundles/api/bundle1/packages.yaml"
+     ],
+     "tests/test_files/bundles/api/bundle1/results/bundle.yaml"),
```
kevinrizza: Same comment as above about limiting prs to changes related to the commit message and purpose.

tests/test_files/invalid_malformed_parser_error.yaml
```diff
@@ -0,0 +1,54 @@
...
+          required:
+          - targetNamespace
+          - packages
+          properties:
+            targetNamespace:
+              type: string
+              description: The namespace where the operators will be enabled
+            packages:
+              type: string
+              description: Comma separated list of operator(s) without spaces which will be enabled in the target namespace
```
kevinrizza: NIT: Can we use a consistent naming style for filenames in this folder?

tests/test_files/invalid_malformed_scanner_error.yaml
```diff
@@ -0,0 +1,54 @@
...
+          required:
+          - targetNamespace
+          - packages
+          properties:
+            targetNamespace:
+              type: string
+              description: The namespace where the operators will be enabled
+            packages:
+              type: string
+              description: Comma separated list of operator(s) without spaces which will be enabled in the target namespace
```
kevinrizza: NIT: Can we use a consistent naming style for filenames in this folder?

operatorcourier/identify.py
scoheb: or at least exit non-zero...in addition, with #27 ... this should be a critical error in the json output. @awgreene

jeremylinlin: @scoheb I have discussed with @kevinrizza. Instead of raising an exception, we will just log error messages to the output. And this will be made consistent in the entire program.

MartinBasti: Please never ever call `exit()` in any python module outside CLI processing. It will cause immediate exit from any app using `operator-courier` as its part

operatorcourier/identify.py
jeremylinlin: This may require changes in multiple function signatures. I instead log all YAMLs that are parsed successfully.

operatorcourier/identify.py
jeremylinlin: I now only exit(1) in the main entry of the CLI, if any error bubbles up to this level.

operatorcourier/cli.py
```diff
@@ -2,11 +2,17 @@
...
 def main():
     """Generate the CLI bits
     """
-    parser = _CliParser()
-    parser.parse()
+    try:
+        parser = _CliParser()
+        parser.parse()
+    except Exception:  # Exception raised to CLI should not be thrown to users,
+        exit(1)        # it should just be captured by logs
```
kevinrizza: I still don't understand why we need to use the `exit()` syscall here. Can't we just print a message and return after printing?

operatorcourier/cli.py
```diff
@@ -32,17 +38,16 @@ def parse(self):
         args = parser.parse_args(sys.argv[1:2])
         if not hasattr(self, args.command):
             parser.error('Unrecognized command')
```
kevinrizza: why remove whitespace here? this seems like noise.

operatorcourier/cli.py
jeremylinlin: What I'm thinking is that
  - we should not make exit call within modules, which we have discussed previously
  - the reason why I didn't just print a message and return is that, by doing this the exit status will be 0. But I want the user to know that the CLI program did not exit correctly.

Any thoughts on that?

operatorcourier/cli.py
jeremylinlin: I'll remove those then. I think my IDE did this, but I'll make sure this doesn't happen in the future.

operatorcourier/cli.py
kevinrizza: Right, this is the cli. This looks fine.

MartinBasti: I'm not avare that argparser is giving traceback for user errors, only for developer errors.

This case just hides unexpected errors, you at least needs
```
except Exception as e:
   sys.exit(str(e))
```

operatorcourier/cli.py
```diff
@@ -2,11 +2,17 @@
 import sys
 from operatorcourier import api
 
+sys.tracebacklimit = 0  # this suppresses stacktrace of the CLI tool in the command line
```
MartinBasti: GL with bug reports without stack traces

operatorcourier/identify.py
```diff
@@ -1,16 +1,34 @@
...
-            return "Package"
-        elif "kind" in operatorArtifact:
-            if operatorArtifact["kind"] in ("ClusterServiceVersion", "CustomResourceDefinition"):
-                return operatorArtifact["kind"]
-    raise ValueError('Courier requires valid CSV, CRD, and Package files')
+    try:
+        operatorArtifact = safe_load(operatorArtifactString)
+    except MarkedYAMLError:
+        logger.error("Courier requires valid input YAML files")
+        raise ValueError('Courier requires valid input YAML files')
```
MartinBasti: Pro tip:
```python
msg = "Courier requires valid input YAML files"
logger.error(msg)
raise ValueError(msg)
```

operatorcourier/identify.py
```diff
@@ -1,16 +1,34 @@
...
+        if operatorArtifact is not None:
+            if "packageName" in operatorArtifact:
+                artifact_type = "Package"
+                artifact_name = operatorArtifact['packageName']
+            elif "kind" in operatorArtifact:
+                if operatorArtifact["kind"] in ("ClusterServiceVersion", "CustomResourceDefinition"):
+                    artifact_type = operatorArtifact["kind"]
+                    artifact_name = operatorArtifact['metadata']['name']
+            if artifact_type is not None:
+                logger.info(f'Parsed {artifact_type}: {artifact_name}')
```
MartinBasti: No format strings (`f'...'`)! It is supported only in py 3.7+. We need older python versions too at least 3.6.

Also this logging is not lazy, please use `logger.info('Parsed %s: %s', artifact_type, artifact_name)` instead.

operatorcourier/identify.py
```diff
@@ -1,16 +1,34 @@
...
 
 def get_operator_artifact_type(operatorArtifactString):
-    """get_operator_artifact_type takes a yaml string and determines if it is one of the 
+    """get_operator_artifact_type takes a yaml string and determines if it is one of the
     expected bundle types: ClusterServiceVersion, CustomResourceDefinition, or Package.
 
     :param operatorArtifactString: Yaml string to type check
     """
-    operatorArtifact = yaml.safe_load(operatorArtifactString)
-    if isinstance(operatorArtifact, dict):
```
MartinBasti: You don't need this check anymore? What if yaml contains only list?

operatorcourier/identify.py
MartinBasti: Ah sorry format strings are supported in py 3.6, but there is still that lazy not lazy logging.

jeremylinlin:
@MartinBasti Thanks for your comments. I have updated the PR.

tests/test_files/invalid.malformed.parser.error.yaml
```diff
@@ -0,0 +1,54 @@
+- apiVersion: apiextensions.k8s.io/v1beta1
```
awgreene: Could you add a comment highlighting where the yaml is malformed? This will help debugging issues in the future.

tests/test_files/invalid.malformed.scanner.error.yaml
```diff
@@ -0,0 +1,54 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
awgreene: Could you add a comment highlighting where the yaml is malformed? This will help debugging issues in the future.

PR #36, opened 2019-03-05
Adding UI validation functions for operatorhub.io

galletti94:
I'm adding the WIP tag because I am adding some tests for the UI, but @kevinrizza you're welcome to start reviewing whenever you have time.

/cc @kevinrizza @jeff-phillips-18 @awgreene @dmesser

### DIFF
--- a/operatorcourier/api.py
+++ b/operatorcourier/api.py
@@ -17,7 +17,7 @@ from operatorcourier.nest import nest_bundles
 logging.basicConfig()
 logger = logging.getLogger(__name__)
 
-def build_and_verify(source_dir=None, yamls=None):
+def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False):
     """Build and verify constructs an operator bundle from a set of files and then verifies it for usefulness and accuracy.
     It returns the bundle as a string.
 
@@ -41,7 +41,7 @@ def build_and_verify(source_dir=None, yamls=None):
 
     bundle = BuildCmd().build_bundle(yaml_files)
 
-    valid = ValidateCmd().validate(bundle)
+    valid = ValidateCmd(ui_validate_io).validate(bundle)
 
     if not valid:
         bundle = None
--- a/operatorcourier/cli.py
+++ b/operatorcourier/cli.py
@@ -39,9 +39,10 @@ These are the commands you can use:
     def verify(self):
         parser = argparse.ArgumentParser(description='Build and verify an operator bundle to test')
         parser.add_argument('source_dir', help='Path of your directory of yaml files to bundle. Either set this or use the files argument for bundle data.')
-        
-        args, leftovers = parser.parse_known_args(sys.argv[2:])        
-        api.build_and_verify(source_dir=args.source_dir)
+        parser.add_argument('--ui_validate_io', help='Validate bundle for operatorhub.io UI', action='store_true')
+
+        args, leftovers = parser.parse_known_args(sys.argv[2:])       
+        api.build_and_verify(source_dir=args.source_dir, ui_validate_io=args.ui_validate_io)
     
     # Parse the push command
     def push(self):
--- a/operatorcourier/const_io.py
+++ b/operatorcourier/const_io.py
@@ -1,0 +1,97 @@
+general_required_fields = [
+    "spec"
+]
+
+metadata_required_fields = [
+    {
+        "field" : "annotations",
+        "description" : "Without this field, your operator will be missing essential data for the UI. Please add this field to the CSV and run validation again to see what subfields are required.",
+        "required" : True
+    }
+]
+
+metadata_annotations_required_fields = [
+    {
+        "field" : "description",
+        "description" : "Without this field, the description displayed in the tiles of the UI will be a truncated version of spec.description.",
+        "required" : True
+    },
+    {
+        "field" : "categories",
+        "description" : "Without this field, the operator will be categorized as Other.",
+        "required" : False
+    },
+    {
+        "field" : "capabilities",
+        "description" : "Without this field, the operator will be assigned the basic install capability - you can read more about operator maturity models here https://www.operatorhub.io/getting-started#How-do-I-start-writing-an-Operator?.",
+        "required" : True
+    },
+    {
+        "field" : "repository",
+        "description" : "Without this field, the link to the operator source code will not be displayed in the UI.",
+        "required" : False
+    },
+    {
+        "field" : "createdAt",
+        "description" : "Without this field, the time stamp at which the operator was created will not be displayed in the UI.",
+        "required" : False
+    },
+    {
+        "field" : "containerImage",
+        "description" : "Without this field, the link to the operator image will not be displayed in the UI.",
+        "required" : True
+    },
+    {
+        "field" : "alm-examples",
+        "description" : "Without this field, users will not have examples of how to write Custom Resources for the operator.",
+        "required" : False
+    }
+]
+
+spec_required_fields = [
+    {
+        "field" : "displayName",
+        "description" : "Without this field, the name of the operator for both the tile and the details view will default to metadata.name.",
+        "required" : True
+    },
+    {
+        "field" : "description",
+        "description" : "Without this field, the description in the details view of the operator will default to metadata.annotations.description.",
+        "required" : True
+    },
+    {
+        "field" : "version",
+        "description" : "Without this field, the version in the details view of the operator will be empty.",
+        "required" : True
+    },
+    {
+        "field" : "links",
+        "description" : "Without this field, no links will be displayed in the details page side panel. You can for example link to some additional Documentation, related Blogs or Repositories.",
+        "required" : True
+    },
+    {
+        "field" : "icon",
+        "description" : "Without this field, the operator will display a default operator framework icon.",
+        "required" : False
+    },
+    {
+        "field" : "provider",
+        "description" : "Without this field, users will not be able to filter for provider.",
+        "required" : True
+    },
+    {
+        "field" : "maintainers",
+        "description" : "Without this field, the operator details page will not display the name and contact for users to get support in using the operator. The field should be a yaml list of name & email pairs.",
+        "required" : True
+    },
+    {
+        "field" : "customresourcedefinitions",
+        "description" : "Without this field, the operator details page will not display any CRDs the operator needs in order to function.",
+        "required" : False
+    },
+    {
+        "field" : "installModes",
+        "description" : "Without this field, OLM will not be able to deploy your operator.",
+        "required" : True
+    }
+]
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -1,5 +1,10 @@
 import logging
 import json
+import re
+
+import validators as v
+
+from .const_io import general_required_fields, metadata_required_fields, metadata_annotations_required_fields, spec_required_fields
 
 logger = logging.getLogger(__name__)
 
@@ -9,7 +14,8 @@ class ValidateCmd():
     csvKey = "clusterServiceVersions"
     pkgsKey = "packages"
 
-    def __init__(self):
+    def __init__(self, ui_validate_io=False):
+        self.ui_validate_io = ui_validate_io
         pass
         
     def validate(self, bundle):
@@ -36,6 +42,8 @@ class ValidateCmd():
         validationDict[self.crdKey] = self._type_validation(bundleData, self.crdKey, self._crd_validation, False)  
         validationDict[self.csvKey] = self._type_validation(bundleData, self.csvKey, self._csv_validation, True)
         validationDict[self.pkgsKey] = self._type_validation(bundleData, self.pkgsKey, self._pkgs_validation, True)
+        if self.ui_validate_io:
+            validationDict[self.csvKey] &= self._type_validation(bundleData, self.csvKey, self._ui_validation_io, True)
 
         valid = True
         for key, value in validationDict.items():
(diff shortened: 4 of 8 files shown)


operatorcourier/validate.py
```diff
@@ -237,3 +243,231 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+            levels = [
+                "Basic Install",
+                "Seamless Upgrades",
+                "Full Lifecycle",
+                "Deep Insight",
+                "Auto Pilot"
+            ]
+            return field in levels
+
+        def is_date(field):
```
galletti94: What's the expected format for date? @jeff-phillips-18 is there a convention we need for our db?

operatorcourier/cli.py
```diff
@@ -39,9 +39,10 @@ def parse(self):
     def verify(self):
         parser = argparse.ArgumentParser(description='Build and verify an operator bundle to test')
         parser.add_argument('source_dir', help='Path of your directory of yaml files to bundle. Either set this or use the files argument for bundle data.')
-        
-        args, leftovers = parser.parse_known_args(sys.argv[2:])        
-        api.build_and_verify(source_dir=args.source_dir)
+        parser.add_argument('--ui_validate', help='Set if you want the yaml files validated for the operatorhub.io UI', action='store_true')
```
SamiSousa: `Validate bundle for operatorhub.io UI`

operatorcourier/const.py
```diff
@@ -0,0 +1,97 @@
...
+        "field" : "customresourcedefinitions",
+        "description" : "Without this field, the operator details page will not display any CRDs the operator needs in order to function.",
+        "required" : False
+    },
+    {
+        "field" : "installModes",
+        "description" : "Without this field, OLM will not be able to deploy your operator.",
+        "required" : True
+    }
+]
```
SamiSousa: nit: newline

operatorcourier/const.py
```diff
@@ -0,0 +1,97 @@
...
+
+metadata_annotations_required_fields = [
+    {
+        "field" : "description",
+        "description" : "Without this field, the description displayed in the tiles of the UI will be a truncated version of spec.description.",
+        "required" : True
+    },
+    {
+        "field" : "categories",
+        "description" : "Without this field, the operator will be categorized as Other.",
```
SamiSousa: Since there is a default, should this be required? What factors decide which fields are "required"?

operatorcourier/const.py
```diff
@@ -0,0 +1,97 @@
+general_required_fields = [
```
SamiSousa: Some of these fields are checked for in the default operator-courier validation. There may be some overlap in these checks and the one's already present.

operatorcourier/validate.py
```diff
@@ -237,3 +245,135 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+        csvs = bundleData[self.csvKey]
+
+        for csv in csvs:
+            if self._ui_csv_fields_exist_validation(csv) is False:
+                logger.error("UI validation failed to verify required fields exist.")
+                valid = False
+            
+            if valid:
+                if self._ui_csv_fields_format_validation(csv) is False:
+                    logger.error("UI validation failed to verify required fields exist.")
```
SamiSousa: Can this error be more descriptive? Getting here requires that fields exist, but isn't formatted properly. Is that correct?

operatorcourier/validate.py
```diff
@@ -237,3 +245,135 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+                return False
+            return True
+
+        def is_email(field):
+            if not v.email(field):
+                return False
+            return True
+
+        def is_version(field):
+            pattern1 = re.compile(r'v(\d+\.)(\d+\.)(\d)')
```
SamiSousa: I believe version also supports things like `-alpha` that can be added to the end of a semver.
Can these patterns be consolidated? Eg: `r'v?(\d+\.)(\d+\.)(\d)(-[a-z]+)?'`

operatorcourier/validate.py
```diff
@@ -237,3 +245,135 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+
+        def is_date(field):
+            return True
+
+        valid = True
+
+        spec = csv["spec"]
+
+        for maintainer in spec["maintainers"]:
+            if "name" not in maintainer and "email" not in maintainer:
```
SamiSousa: Are both "name" and "email" required? If so, this should use `or`.

operatorcourier/validate.py
```diff
@@ -237,3 +245,135 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+                logger.error("csv.spec.maintainers element should contain both name and email")
+                valid = False
+            else:
+                if not is_email(maintainer["email"]):
+                    logger.error("%s is not a valid email", maintainer["email"])
+                    valid = False
+        
+        for link in spec["links"]:
+            if "name" not in link and "url" not in link:
```
SamiSousa: This should also be `or` rather than `and` if both are required

operatorcourier/validate.py
```diff
@@ -237,3 +245,135 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+
+        if not is_capability_level(annotations["capabilities"]):
+            logger.error("metadata.annotations.capabilities %s is not a valid capabilities level", annotations["capability"])
+            valid = False
+        
+        if not is_date(annotations["createdAt"]):
+            logger.error("metadata.annotations.createdAt %s is not a valid date", annotations["createdAt"])
+            valid = False
+
+        return valid
```
SamiSousa: nit: newline

operatorcourier/const.py
galletti94: Your (everyone added as a reviewer) feedback in this PR determines which fields are required

operatorcourier/validate.py
galletti94: This is the general error that will be displayed after the validation function returns. The exact errors that caused validation to fail will be listed before the above error.

operatorcourier/validate.py
galletti94: Good point. @jeff-phillips-18 can you confirm that this is allowed for the UI?

operatorcourier/const.py
galletti94: Yeah I realize that but I'm not sure how to bypass that in a way that the default validation and ui validation stay in sync. I'm open to suggestions but IMO a little overlap could be better than more complexity

operatorcourier/const.py
awgreene: Let's have this conversation before I review. In my opinion, `required` in this case should be limited to issues that would prevent the UI from displaying as intended. Anything that has a default should not be `required`.

galletti94: The thing is we have defaults out of necessity due to there being no enforcement on the csv for ui fields. Because operatorhub.io is in production we have had to implement fallbacks for every scenario because we can't have the UI failing. I agree we should have a discussion about what is required but it should not be on the basis of what would make the UI fail and should rather focus on what fields are required for the user experience to make sense.

operatorcourier/validate.py
SamiSousa: Actually it may be better to check with this package: https://pypi.org/project/semver/
We should use a regex that matches what quay.io is using to be complete (the above isn't that :/).

operatorcourier/cli.py
```diff
@@ -39,9 +39,11 @@ def parse(self):
     def verify(self):
         parser = argparse.ArgumentParser(description='Build and verify an operator bundle to test')
         parser.add_argument('source_dir', help='Path of your directory of yaml files to bundle. Either set this or use the files argument for bundle data.')
-        
-        args, leftovers = parser.parse_known_args(sys.argv[2:])        
-        api.build_and_verify(source_dir=args.source_dir)
+        parser.add_argument('--ui_validate_io', help='Validate bundle for operatorhub.io UI', action='store_true')
+        parser.add_argument('--ui_validate_ocp', help='Validate bundle for OCP Operator Hub UI', action='store_true')
```
SamiSousa: I believe this should be `OperatorHub`, since there is a pending PR in console changing the name to this officially: https://github.com/openshift/console/pull/1192

operatorcourier/const_io.py
```diff
@@ -0,0 +1,97 @@
...
+        "required" : True
+    },
+    {
+        "field" : "version",
+        "description" : "Without this field, the version in the details view of the operator will be empty.",
+        "required" : True
+    },
+    {
+        "field" : "links",
+        "description" : "Without this field, no links will be displayed in the details page side panel.",
```
SamiSousa: Could this description include the suggestion for a "Documentation" link?

operatorcourier/const_io.py
```diff
@@ -0,0 +1,97 @@
...
+        "required" : False
+    },
+    {
+        "field" : "provider",
+        "description" : "Without this field, users will not be able to filter for provider.",
+        "required" : True
+    },
+    {
+        "field" : "maintainers",
+        "description" : "Without this field, the operator details page will not display the name and contact for users to get support in using the operator.",
```
SamiSousa: Could this include an example of what the field should look like? eg:
```
  maintainers:
  - name: Example
    email: example@email.com
```
There should be some mention of CSV documentation on a validation failure or warning to help people figure out how to populate these fields.

operatorcourier/validate.py
```diff
@@ -237,3 +248,152 @@ def _type_validation(self, bundleData, typeName, validator, required):
             valid = False
 
         return valid
+
+    def _ui_validation_ocp(self, bundleData):
+        valid = True
+        logger.info("Validating cluster service versions for OCP Operator Hub UI.")
```
SamiSousa: This looks like a placeholder, will this be implemented in a future PR?

operatorcourier/validate.py
```diff
@@ -237,3 +248,152 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+        annotations = csv["metadata"]["annotations"]
+        
+        if "customresourcedefinitions" in spec:
+            if "owned" in spec["customresourcedefinitions"]:
+                crds = spec["customresourcedefinitions"]["owned"]
+                if "alm-examples" in annotations:
+                    alm_kinds = get_alm_kinds(json.loads(annotations["alm-examples"]))
+                    for crd in crds:
+                        if crd["kind"] not in alm_kinds:
+                            logger.error("%s CRD does not have an entry in alm-examples - please add such an example CR.", crd["kind"])
```
SamiSousa: If `alm-examples` is not required, should the field set valid to False if a CRD example is missing?

operatorcourier/validate.py
```diff
@@ -237,3 +248,152 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+                            valid = False
+                else:
+                    logger.error("You should have alm-examples for every owned CRD")
```
SamiSousa: Are we enforcing this?

operatorcourier/validate.py
```diff
@@ -237,3 +248,152 @@ def _type_validation(self, bundleData, typeName, validator, required):
...
+                    valid = False
+
+        for maintainer in spec["maintainers"]:
```
SamiSousa: There should also be some validation for `spec.provider` as well. It should check that `spec.provider` is set to an object with key "name" and a string value.

tests/test_files/bundles/verification/valid.bundle.yaml
```diff
@@ -4,13 +4,47 @@ data:
       kind: ClusterServiceVersion
       metadata:
         annotations:
-          alm-examples: '[{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdCluster","metadata":{"name":"example","namespace":"default"},"spec":{"size":3,"version":"3.2.13"}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdRestore","metadata":{"name":"example-etcd-cluster"},"spec":{"etcdCluster":{"name":"example-etcd-cluster"},"backupStorageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdBackup","metadata":{"name":"example-etcd-cluster-backup"},"spec":{"etcdEndpoints":["<etcd-cluster-endpoints>"],"storageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}}]'
+          alm-examples: |-
```
SamiSousa: Thank you for the multi-line JSON :+1:

tests/test_files/bundles/verification/valid.bundle.yaml
```diff
@@ -4,13 +4,47 @@ data:
...
+                    }
+                }
+              
+            ]
           categories: openshift required
           certified: 'true'
           containerImage: quay.io/openshift/origin-operator-marketplace:latest
           createdAt: 2019/11/15
           description: An operator to run the OpenShift marketplace
-          healthIndex: B
```
SamiSousa: Does the courier have a check for this field? If so, it should be removed.

operatorcourier/validate.py
galletti94: That's the plan

operatorcourier/validate.py
galletti94: it's not required because `customresourcedefinitions` is not required. If `customresourcedefinitions` is present then `alm-examples` is required

operatorcourier/validate.py
galletti94: Yes

tests/test_files/bundles/verification/valid.bundle.yaml
galletti94: I don't think it does

operatorcourier/api.py
```diff
@@ -17,7 +17,7 @@
 logging.basicConfig()
 logger = logging.getLogger(__name__)
 
-def build_and_verify(source_dir=None, yamls=None):
+def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False, ui_validate_ocp=False):
```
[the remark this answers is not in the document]
galletti94: could be. I'll wait to see what others say before changing this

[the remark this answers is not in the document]
dmesser: you could also simply call it `ui_validate_openshift`

galletti94:
Removed the WIP tag. @awgreene @SamiSousa @dmesser @kevinrizza @aravindhp @jeff-phillips-18 @anik120 @jeremylinlin Please review

operatorcourier/api.py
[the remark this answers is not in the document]
kevinrizza: +1 can this be `ui_validate_openshift`? I don't think this is ocp specific.

operatorcourier/cli.py
```diff
@@ -39,9 +39,11 @@ def parse(self):
     def verify(self):
         parser = argparse.ArgumentParser(description='Build and verify an operator bundle to test')
         parser.add_argument('source_dir', help='Path of your directory of yaml files to bundle. Either set this or use the files argument for bundle data.')
-        
-        args, leftovers = parser.parse_known_args(sys.argv[2:])        
-        api.build_and_verify(source_dir=args.source_dir)
+        parser.add_argument('--ui_validate_io', help='Validate bundle for operatorhub.io UI', action='store_true')
+        parser.add_argument('--ui_validate_ocp', help='Validate bundle for OCP OperatorHub UI', action='store_true')
```
kevinrizza: Same comment as above, can this flag be `--ui_validate_openshift` ?

operatorcourier/validate.py
```diff
@@ -9,7 +14,9 @@ class ValidateCmd():
     csvKey = "clusterServiceVersions"
     pkgsKey = "packages"
 
-    def __init__(self):
+    def __init__(self, ui_validate_io=False, ui_validate_ocp=False):
```
kevinrizza: same comment about flag naming

operatorcourier/validate.py
```diff
@@ -237,3 +248,179 @@ def _type_validation(self, bundleData, typeName, validator, required):
             valid = False
 
         return valid
+
+    def _ui_validation_ocp(self, bundleData):
+        valid = True
+        logger.info("Validating cluster service versions for OCP OperatorHub UI.")
```
kevinrizza: is this intended to just do nothing right now?

galletti94: @kevinrizza yeah is that ok? Intending to add / refactor for common code in a future PR. Happy to remove this for now if you think that's better.

kevinrizza: @galletti94 I think if it does nothing right now we should just omit it. We can come back later and create it when we implement it. Otherwise it could give a false positive.

operatorcourier/validate.py
jeff-phillips-18: Yes that would work for the UI. We only really use it to sort the different versions, for version selection dropdown and for determining the latest version.

Issue #39, opened 2019-03-06
Consider using JSON schema for validation


csomh:
YAML can easily be transformed to JSON, and there is [JSON schema](https://json-schema.org/), which can be used to describe how a valid artifact should look like. There is also a [Python package](https://pypi.org/project/jsonschema/) that implements the specification and provides validation.

Would it make sense to follow this path for validating artifacts, instead of writing and testing a lot of validation code? Schema(s) will need testing, but the effort would be much lower.

scoheb:
A nice example of this can be found here: https://pagure.io/fedora-ci/messages/

kevinrizza:
The problem with schema testing is that many layers of edge cases for these bundles. In many cases, it is not simple enough for us to verify that certain fields exist or not and what their types are. We have more complex validation logic that verifies that, if a field exists in one file, that a corresponding field or value exists in another file. For example:

Iterate over the `csv`'s `customresourcedefinitions.owned` list, and for each one ensure that, in the list of `crd` files, ensure that one and exactly one's `metadata.name` matches `customresourcedefinitions.owned.name`

I think that use cases like that can't really be defined by a simple schema validation. @scoheb @csomh please correct me if I am missing or misunderstanding something here.

csomh:
@kevinrizza, yes, you are right that jsonschema might not be able to check such constraints, without some external help. Still, reading through `validate.py` and `const_io.py`, it seems that most of the checks currently done by Operator Courier are actually not of this kind. 

Instead, the code mostly checks if fields are present or have the right type. And implements some ways of reporting validation errors. All these have well tested implementations in jsonschema.

Let me give you an example: `ValidateCmd._pkgs_validation()` could be replaced by the following piece of schema:
```json
{
    "title": "Package manifest", 
    
    "type": "object",
    
    "properties": {
        "packageName": {
            "type": "string",
            "description": "Name of the package"
        },
        "channels": {
            "type": "array",
            "items": {
                "type": "object",
                "properties": {
                    "name": {
                        "type": "string"
                    },
                    "currentCSV": {
                        "type": "string"
                    }
                },
                "required": [
                    "name",
                    "currentCSV"
                ]
            },
            "minItems": 1,
            "uniqueItems": true
        }
    },
    "required": [
        "packageName",
        "channels"
    ]
}
```

The above schema does not check if `currentCSV` is one of the CSV names, but should be trivial to create a sub-schema for this on the fly.

[...]

Issue #17, opened 2019-02-14
Improve how yaml paths are vertified


awgreene:
Multiple if statements are currently used to validate dictionary fields. I recommend creating a method that:

1. Accepts a `dict` and a path (example path: `metadata.annotations.alm-examples`)
2. Checks that the path exists within the `dict`
3. Returns true/false

PR #41, opened 2019-03-07
Check presence and match of fields in CRD and CSV in bundle

jeremylinlin:
- Check if the following fields in CRD matches that in CSV in bundle
  - `CRD.spec.names.plural`.`CRD.spec.group` == `CSV.spec.crd.owned.name`
  - `CRD.spec.names.kind`                    == `CSV.spec.crd.owned.kind`
  - `CRD.spec.version`                       == `CSV.spec.crd.owned.version`
- Check the presence of all above fields in the bundle

### DIFF
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -107,6 +107,23 @@ class ValidateCmd():
             if "spec" not in crd:
                 self._log_error("crd spec not defined.")
                 valid = False
+            else:
+                if "names" not in crd['spec']:
+                    self._log_error("crd spec.names not defined.")
+                    valid = False
+                else:
+                    if "kind" not in crd['spec']['names']:
+                        self._log_error("crd spec.names.kind not defined.")
+                        valid = False
+                    if "plural" not in crd['spec']['names']:
+                        self._log_error("crd spec.names.plural not defined.")
+                        valid = False
+                if "group" not in crd['spec']:
+                    self._log_error("crd spec.group not defined.")
+                    valid = False
+                if "version" not in crd['spec']:
+                    self._log_error("crd spec.version not defined.")
+                    valid = False
 
         return valid
 
@@ -169,18 +186,66 @@ class ValidateCmd():
             if "owned" not in customresourcedefinitions:
                 self._log_error("spec.customresourcedefinitions.owned"
                                 "not defined for csv")
-                valid = False
-            else:
-                for crd in customresourcedefinitions["owned"]:
-                    if "name" not in crd:
-                        self._log_error("name not defined for item in "
-                                        "spec.customresourcedefinitions.")
-                        valid = False
-                    else:
-                        if crd["name"] not in crdList:
-                            self._log_error("custom resource definition referenced "
-                                            "in csv not defined in root list of crds")
+                return False
 
+            for csvOwnedCrd in customresourcedefinitions["owned"]:
+                if "name" not in csvOwnedCrd:
+                    self._log_error("name not defined for item in "
+                                    "spec.customresourcedefinitions.")
+                    valid = False
+                elif csvOwnedCrd["name"] not in crdList:
+                    self._log_error("custom resource definition %s referenced in csv "
+                                    "not defined in root list of crds",
+                                    csvOwnedCrd["name"])
+                    valid = False
+
+                if "kind" not in csvOwnedCrd:
+                    self._log_error("kind not defined for item in "
+                                    "spec.customresourcedefinitions.")
+                    valid = False
+                if "version" not in csvOwnedCrd:
+                    self._log_error("version not defined for item in "
+                                    "spec.customresourcedefinitions.")
+                    valid = False
+
+                for crd in bundleData[self.crdKey]:
+                    if 'name' not in csvOwnedCrd:
+                        continue
+                    if 'metadata' not in crd or 'name' not in crd['metadata']:
+                        continue
+                    if csvOwnedCrd['name'] != crd['metadata']['name']:
+                        continue
+
+                    if 'kind' in csvOwnedCrd:
+                        if 'spec' in crd:
+                            if 'names' in crd['spec']:
+                                if 'kind' in crd['spec']['names']:
+                                    if csvOwnedCrd['kind'] != \
+                                            crd['spec']['names']['kind']:
+                                        self._log_error('CRD.spec.names.kind does not '
+                                                        'match CSV.spec.crd.owned.kind')
+                                        valid = False
+
+                    if 'version' in csvOwnedCrd:
+                        if 'spec' in crd:
+                            if 'version' in crd['spec']:
+                                if csvOwnedCrd['version'] != crd['spec']['version']:
+                                    self._log_error('CRD.spec.version does not match '
+                                                    'CSV.spec.crd.owned.version')
+                                    valid = False
+
+                    if 'name' in csvOwnedCrd:
+                        if 'spec' in crd:
+                            if 'names' in crd['spec'] and 'group' in crd['spec']:
+                                if 'plural' in crd['spec']['names']:
+                                    if csvOwnedCrd['name'] != \
+                                            crd['spec']['names']['plural'] + '.' + \
+                                            crd['spec']['group']:
+                                        self._log_error("`CRD.spec.names.plural`."
+                                                        "`CRD.spec.group` does not "
+                                                        "match "
+                                                        "CSV.spec.crd.owned.name")
+                                        valid = False
         return valid
 
     def _csv_metadata_validation(self, metadata):
--- a/tests/test_files/bundles/verification/crdmissingkindfield.invalid.bundle.yaml
+++ b/tests/test_files/bundles/verification/crdmissingkindfield.invalid.bundle.yaml
@@ -1,0 +1,301 @@
+# CRD.spec.names.kind in one of the CRDs is missing in this file
+data:
+  clusterServiceVersions: |
+    - apiVersion: operators.coreos.com/v1alpha1
+      kind: ClusterServiceVersion
+      metadata:
+        annotations:
+          alm-examples: '[{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdCluster","metadata":{"name":"example","namespace":"default"},"spec":{"size":3,"version":"3.2.13"}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdRestore","metadata":{"name":"example-etcd-cluster"},"spec":{"etcdCluster":{"name":"example-etcd-cluster"},"backupStorageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}},{"apiVersion":"etcd.database.coreos.com/v1beta2","kind":"EtcdBackup","metadata":{"name":"example-etcd-cluster-backup"},"spec":{"etcdEndpoints":["<etcd-cluster-endpoints>"],"storageType":"S3","s3":{"path":"<full-s3-path>","awsSecret":"<aws-secret>"}}}]'
+          categories: openshift required
+          certified: 'true'
+          containerImage: quay.io/openshift/origin-operator-marketplace:latest
+          createdAt: 2019/11/15
+          description: An operator to run the OpenShift marketplace
+          healthIndex: B
+          repository: https://github.com/operator-framework/operator-marketplace
+          support: Red Hat
+        name: marketplace-operator.v0.0.1
+        namespace: placeholder
+      spec:
+        customresourcedefinitions:
+          owned:
+          - description: Represents an OperatorSource.
+            displayName: Operator Source
+            kind: OperatorSource
+            name: operatorsources.marketplace.redhat.com
+            specDescriptors:
+            - description: The type of the operator source.
+              displayName: Type
+              path: type
+            - description: Points to the remote app registry server from where operator
+                manifests can be fetched.
+              displayName: Endpoint
+              path: endpoint
+            - description: 'The namespace in app registry.
+
+                Only operator manifests under this namespace will be visible.
+
+                Please note that this is not a k8s namespace.'
+              displayName: Registry Namespace
+              path: registryNamespace
+            statusDescriptors:
+            - description: Current status of the CatalogSourceConfig
+              displayName: Current Phase Name
+              path: currentPhase.phase.name
+            - description: Message associated with the current status
+              displayName: Current Phase Message
+              path: currentPhase.phase.message
+            version: v1alpha1
+          - description: Represents a CatalogSourceConfig object which is used to configure
+              a CatalogSource.
+            displayName: Catalog Source Config
+            kind: CatalogSourceConfig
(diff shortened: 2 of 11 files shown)


operatorcourier/validate.py
```diff
@@ -128,16 +128,33 @@ def _csv_spec_validation(self, spec, bundleData):
...
-                        logger.error("name not defined for item in spec.customresourcedefinitions.")
-                        valid = False
-                    else:
-                        if crd["name"] not in crdList:
-                            logger.error("custom resource definition referenced in csv not defined in root list of crds")
-
+                return False
+
+            for csvOwnedCrd in customresourcedefinitions["owned"]:
+                if "name" not in csvOwnedCrd:
```
awgreene: Should we create a list like ["name", "kind", ..] and merge these if statements into a forloop with an if statement?

Thoughts @kevinrizza @jeremylinlin

tests/test_files/bundles/verification/crdmissingkindfield.invalid.bundle.yaml
```diff
@@ -0,0 +1,301 @@
+# CRD.spec.names.kind in one of the CRDs is missing in this file
```
awgreene: There have been a number of large yaml files for tests added recently. Should we create some smaller yamls templates that could be modified to test specific errors? @kevinrizza @jeremylinlin

operatorcourier/validate.py
jeremylinlin: I created this [helper](https://github.com/operator-framework/operator-courier/pull/41/files/a9b0bbdf052014dd5ce4a696e3d1f15f69cdf8a3#diff-4badd63917dd32ab713b1c89ddfa260cR52) that recursively checks the presence of [required nested fields](https://github.com/operator-framework/operator-courier/pull/41/files#diff-4badd63917dd32ab713b1c89ddfa260cR11) in yaml files. It should work for both CRDs and CSVs. Maybe I should do it in a separate PR for [this issue](https://github.com/operator-framework/operator-courier/issues/17)? @awgreene

tests/test_files/bundles/verification/crdmissingkindfield.invalid.bundle.yaml
kevinrizza: @awgreene my preference is a separate yaml for each individual negative test. is there a reason to make them smaller?

operatorcourier/validate.py
```diff
@@ -24,52 +32,44 @@ def validate(self, bundle):
...
             logger.error("Bundle does not contain base data field.")
             #break here because there's nothing else to learn
             return False
 
         bundleData = bundle[self.dataKey]
 
-        validationDict[self.crdKey] = self._type_validation(bundleData, self.crdKey, self._crd_validation, False)  
-        validationDict[self.csvKey] = self._type_validation(bundleData, self.csvKey, self._csv_validation, True)
-        validationDict[self.pkgsKey] = self._type_validation(bundleData, self.pkgsKey, self._pkgs_validation, True)
+        if not self._type_validation(bundleData, self.crdKey, self._crd_validation, False):
```
kevinrizza: why refactor this? right now it is set up specifically so that we can get the range of all errors, rather than fail as soon as one of these validation tests runs.

jeremylinlin: @kevinrizza oh I didn't know it was intentional. I did it for a reason. Because when I check the fields match in csv spec validation, I would expect it passes all the CRD presence check so that I will directly access the field and compare. If some fields are missing then [this line](https://github.com/operator-framework/operator-courier/pull/41/files#diff-4badd63917dd32ab713b1c89ddfa260cR149) will raise an exception which accesses a non-existing field.

kevinrizza: @jeremylinlin Just reordering those checks would cause your code to fail suddenly. Let's be more careful and when you are accessing fields that could not exist inside the map always check for them to exist first. We should consider each of the checks in the validator to be independent and not come with any previous knowledge. While it's true that it will reduce performance somewhat, we are not really performance bound by these checks right now.

operatorcourier/validate.py
```diff
@@ -85,30 +93,27 @@ def _bundle_validation(self, bundle):
...
-                self._log_error("crd apiVersion not defined.")
-                valid = False
+            for crd_required_nested_field in self.crd_required_nested_fields:
+                valid &= self._crd_fields_validation(crd_required_nested_field.split('.'), crd, "")
+        return valid
 
-            if "spec" not in crd:
-                self._log_error("crd spec not defined.")
-                valid = False
+    # TODO: refactor csv and pkg fields presence validations with this helper
```
kevinrizza: We shouldn't leave TODOs in our code unless it is something that is important for the documentation of the source code itself (for example, if what was committed was a necessary workaround that requires understanding later). In this case, this TODO just seems like another work item we should do later. Please omit it.

tests/test_files/bundles/verification/crdmissingkindfield.invalid.bundle.yaml
awgreene: I would prefer that we have a `tests/templates` dir that contains a few "perfect" CSVs/CRDS/Packages which are copied to a dict and edited to contain the errors that the tests are specifically looking for.

If we continue to use a number of different CSVs, The tests we write will become increasingly polluted as the validation tests are added. If the tests aren't polluted, that means the CSVs will be updated to account for the validation changes. In this case, a user may have to update a number of CSVs to introduce a new validation test. This becomes increasingly difficult to maintain as more tests are added.

jeremylinlin: @awgreene Do you mean we should only load from the "perfect" CSV, programmatically edit the content str after reading the file to introduce the error for the specific test case?

@kevinrizza Thoughts?

kevinrizza: @awgreene @jeremylinlin 

While I can appreciate the desire to avoid having to edit multiple files, I disagree with the notion that we should write complex code that reads a template file and then modifies it inline (especially for yaml as complex as `csv` files). If that code is anything more than completely trivial, then it becomes necessary to test the test code as well. Which, to me, entirely defeats the purpose of unit tests. It may be slightly time consuming to deal with, but it is significantly simpler in complexity to update a bunch of yaml files with a new field if a verification test is added.

operatorcourier/validate.py
```diff
@@ -85,30 +93,27 @@ def _bundle_validation(self, bundle):
...
-                    valid = False
-            else:
-                self._log_error("crd metadata not defined.")
-                valid = False
-
-            if "apiVersion" not in crd:
-                self._log_error("crd apiVersion not defined.")
-                valid = False
+            for crd_required_nested_field in self.crd_required_nested_fields:
+                valid &= self._crd_fields_validation(crd_required_nested_field.split('.'),
```
SamiSousa: nit: Any benefit of using this oneliner over an if statement? It's already long enough to spread two lines

operatorcourier/validate.py
```diff
@@ -85,30 +93,27 @@ def _bundle_validation(self, bundle):
...
+        if len(fields) == 0:
+            return True
+        root_field, subfields = fields[0], fields[1:]
+        field_string = root_field if prefix == "" else prefix + "." + root_field
 
-        return valid
+        if root_field == 'metadata' and 'name' in yaml_dict['metadata']:
+            logging.info("Evaluating crd %s", yaml_dict['metadata']['name'])
+        if root_field not in yaml_dict:
+            self._log_error("crd %s not defined.", field_string)
```
SamiSousa: nit: Can this also print the name of the crd in question? Or is that covered by previous if statement?

operatorcourier/validate.py
```diff
@@ -167,20 +172,54 @@ def _csv_spec_validation(self, spec, bundleData):
...
-                        self._log_error("name not defined for item in "
-                                        "spec.customresourcedefinitions.")
-                        valid = False
-                    else:
-                        if crd["name"] not in crdList:
-                            self._log_error("custom resource definition referenced "
-                                            "in csv not defined in root list of crds")
+                return False
+
+            for csvOwnedCrd in customresourcedefinitions["owned"]:
```
SamiSousa: nit: Are we sticking to a convention on variable and function naming? If so, what is it? Snake case for functions and CamelCase for variables, etc?

operatorcourier/validate.py
```diff
@@ -167,20 +172,54 @@ def _csv_spec_validation(self, spec, bundleData):
...
+                if "name" not in csvOwnedCrd:
+                    self._log_error("name not defined for item in "
+                                    "spec.customresourcedefinitions.")
+                    valid = False
+                elif csvOwnedCrd["name"] not in crdList:
+                    self._log_error("custom resource definition referenced in csv "
```
SamiSousa: Can this print the name of the CRD that doesn't match?

SamiSousa: Can the CRD name also be added to the other error logs here as well?

operatorcourier/validate.py
```diff
@@ -167,20 +172,54 @@ def _csv_spec_validation(self, spec, bundleData):
...
+                    self._log_error("kind not defined for item in "
+                                    "spec.customresourcedefinitions.")
+                    valid = False
+                if "version" not in csvOwnedCrd:
+                    self._log_error("version not defined for item in "
+                                    "spec.customresourcedefinitions.")
+                    valid = False
+
+                for crd in bundleData[self.crdKey]:
+                    if 'name' not in csvOwnedCrd or not \
```
SamiSousa: nit: This if statement is kinda long, and I'm not sure what it's checking for: Are we continuing if the name doesn't exist or if it does exist?

operatorcourier/validate.py
jeremylinlin: Actually using an if statement will also make the line exceed 90 chars. I think the line is long mainly because the long and descriptive method names.

operatorcourier/validate.py
jeremylinlin: > Or is that covered by previous if statement?

Yes.

> Can this also print the name of the crd in question?

This actually prints out the (nested) field that is missing in the CRD. See this [test case](https://github.com/operator-framework/operator-courier/pull/41/files#diff-7488fbee800ac1a3edef38842281bda6R66) as an example. I have updated the log message to make it less confusing.

operatorcourier/validate.py
jeremylinlin: I think this is something we need to fix in the future. Because I have seen both styles in our code, so for now I just pick the style that the code nearby is using.

operatorcourier/validate.py
jeremylinlin: Actually I didn't change much logic in those if statements except making the code less indented.

> Can this print the name of the CRD that doesn't match?

Sure I can do that.

> Can the CRD name also be added to the other error logs here as well?

I can but that will require some additional nested `if` checks to make sure `csvOwnedCrd["name"]` exists and is valid. Do you want me to do that?

operatorcourier/validate.py
jeremylinlin: We continue to check field matches only if the name in both CRD and CSV exists, and they match.

operatorcourier/validate.py
csomh: Please be aware, that `&=` is a bitwise operator. But it seems this code intends to do a boolean operation, so probably

```python
valid = valid and crd_field_valid
```
would better express the intention here.

jeremylinlin: Hi @csomh, thanks for the comment. I have updated this.

operatorcourier/validate.py
SamiSousa: Fair enough

operatorcourier/validate.py
```diff
@@ -167,20 +172,55 @@ def _csv_spec_validation(self, spec, bundleData):
...
+                        self._crd_fields_validation(['metadata', 'name'], crd, '') \
```
SamiSousa: If a CRD in the bundle doesn't have a name, should that also log as an error?

jeremylinlin: Hi @SamiSousa, thanks for the review. Yes, but I believe this should be covered in [previous checks](https://github.com/operator-framework/operator-courier/blob/25af109b1a25a74ab805e489be9981ee413ada9a/operatorcourier/validate.py#L180-L183).

SamiSousa: That check is for the CRD in the CSV, rather than the CRD from the bundle directly.
https://github.com/operator-framework/operator-courier/blob/25af109b1a25a74ab805e489be9981ee413ada9a/operatorcourier/validate.py#L166
On this line the `crdList` is created. Should this be used in this for loop?

kevinrizza: Yeah, for each type we should have a list of fields we check. The fact that they happen to be checked in other places by code that compares those fields to other fields shouldn't be relevant. I think part of the problem is that adding this required fields list `crd_required_nested_fields` makes the code hard to read.

Perhaps instead we should either do that everywhere for every type, or not create that abstraction at all. Either way, it actually seems outside the scope of this pr. Can we revert it?

jeremylinlin: @kevinrizza Without this abstraction we will need to introduce even more nested `if`s just to check the presence of nested fields. These are already checked in CRD checks but we need to do it again to make sure the field in the yaml we compare against csvOwnedCrd actually exists.

Maybe after this PR I can apply this abstraction for checks in all types? And this hopefully can reduce the nested levels like [this](https://github.com/operator-framework/operator-courier/blob/25af109b1a25a74ab805e489be9981ee413ada9a/operatorcourier/validate.py#L276-L307) that can be found in multiple places in this file.

kevinrizza: Your intuition about nested `if`s is good in the general case, but I think in this case it makes the code more clear. If we abstract away certain steps but only in certain cases, it makes the code hard to reason about.

Additionally, I think ultimately implementing a change like the one described here is probably the better long term approach: https://github.com/operator-framework/operator-courier/issues/39

jeremylinlin: @kevinrizza That seems really cool. For this particular PR, what are the changes you want me to make? Should I remove this abstraction and replace it with nested `if`s instead?

kevinrizza: @jeremylinlin yes, and please ensure that any of the checks were doing before are no longer removed.

jeremylinlin: /cc @kevinrizza @SamiSousa Updated.

operatorcourier/validate.py
```diff
@@ -169,18 +186,66 @@ def _csv_spec_validation(self, spec, bundleData):
...
 
+                for crd in bundleData[self.crdKey]:
+                    if 'name' not in csvOwnedCrd:
+                        continue
+                    if 'metadata' not in crd or 'name' not in crd['metadata']:
+                        continue
+                    if csvOwnedCrd['name'] != crd['metadata']['name']:
+                        continue
+
+                    if 'kind' in csvOwnedCrd:
```
SamiSousa: This is much easier to read. :nerd_face:

Issue #24, opened 2019-02-18
Add --version (-v) argument to command line app


scoheb:
Would be nice to know which version is installed.

Avni-Sharma:
PTAL #33

PR #44, opened 2019-03-07
Add --version (-v) argument to CLI

jeremylinlin:
- This commit fixes https://github.com/operator-framework/operator-courier/issues/24

### DIFF
--- a/operatorcourier/cli.py
+++ b/operatorcourier/cli.py
@@ -1,6 +1,8 @@
 import argparse
+import pkg_resources
 import sys
 from operatorcourier import api
+
 
 def main():
     """Generate the CLI bits
@@ -30,7 +32,14 @@ These are the commands you can use:
     push        Create a bundle, test it, and push it to an app registry.
     nest        Take a flat to-be-bundled directory and version nest it.
 ''')
+        try:
+            __version__ = pkg_resources.get_distribution('operator-courier').version
+        except Exception:
+            __version__ = 'unknown'
+
         parser.add_argument('command', help='Subcommand to run')
+        parser.add_argument('-v', '--version', help='Show the current version of operator-courier',
+                            action='version', version=__version__)
 
         args = parser.parse_args(sys.argv[1:2])
         if not hasattr(self, args.command):


operatorcourier/cli.py
```diff
@@ -16,7 +17,12 @@ class _CliParser():
     """
 
     def __init__(self):
-        pass
+        # Several different ways of maintaining a single version number for the project
+        # https://packaging.python.org/guides/single-sourcing-package-version/
```
awgreene: Do we really need to include this comment? We've chosen the method to manage the version.

operatorcourier/cli.py
```diff
@@ -59,7 +65,7 @@ def push(self):
         args, leftovers = parser.parse_known_args(sys.argv[2:])
         api.build_verify_and_push(args.namespace, args.repository, args.release, args.token, source_dir=args.source_dir)
 
-    # Parse the push command
+    # Parse the nest command
```
awgreene: Please make this change in a separate commit.

operatorcourier/cli.py
```diff
@@ -1,6 +1,7 @@
 import argparse
 import sys
 from operatorcourier import api
+import re
```
kevinrizza: can we order imports like:

1. stdlib
2. third party
3. package dependencies

operatorcourier/cli.py
```diff
@@ -16,7 +17,10 @@ class _CliParser():
     """
 
     def __init__(self):
-        pass
+        with open('setup.py', 'r') as f:
```
kevinrizza: shouldn't we only do this in the case that the -v command is specified? rather than during the init

operatorcourier/cli.py
```diff
@@ -19,6 +21,11 @@ def __init__(self):
         pass
 
     def parse(self):
+        def version():
+            with open('setup.py', 'r') as f:
```
MartinBasti: This won't work, `setup.py` is not a part of installed package, this works only from repository locally.

```bash
$ virtualenv venv-3
$ source venv-3/bin/activate
$ pip install git+https://github.com/jeremylinlin/operator-courier.git@add-version-arg-to-cli
$ operator-courier -v                                                                                                    
[Errno 2] No such file or directory: 'setup.py'
```

MartinBasti: Please consult this: https://packaging.python.org/guides/single-sourcing-package-version/

operatorcourier/cli.py
```diff
@@ -1,7 +1,10 @@
 import argparse
+import pkg_resources
 import sys
+import re
```
MartinBasti: unused import

jeremylinlin:
> Thank you for addressing my comments!

Thank you for your very useful comments as well!

Issue #46, opened 2019-03-08
operator-courier should return a non-zero exit code upon ERRORs


dmesser:
Running a validation that procudes `WARNING` and `ERROR`:

```sh
$ operator-courier verify --ui_validate_io strimzi-kafka-operator
WARNING:operatorcourier.validate:csv metadata.annotations.repository not defined. Without this field, the link to the operator source code will not be displayed in the UI.
ERROR:operatorcourier.validate:csv metadata.annotations.capabilities not defined. Without this field, the operator will be assigned the basic install capability - you can read more about operator maturity models here https://www.operatorhub.io/getting-started#How-do-I-start-writing-an-Operator?.
WARNING:operatorcourier.validate:csv metadata.annotations.repository not defined. Without this field, the link to the operator source code will not be displayed in the UI.
ERROR:operatorcourier.validate:UI validation failed to verify required fields for operatorhub.io exist.
ERROR:operatorcourier.api:Bundle failed validation.
```

We should have the exit code of `operator-courier` reflect that.

Current implementation:

```sh
$ echo $?
0
```

kevinrizza:
Looks like this is specifically an issue in the `build_and_verify` method, and consequently the `validate` command in the CLI. Should have a fix up shortly.

PR #47, opened 2019-03-08
Raise exception for validate

kevinrizza:
Problem:
Currently, an exception is not raised if the `validate` method is
called, even when there are errors and the validation fails.
Consequently, the cli module does not catch an exception and the courier
does not modify its exit code from 0 when the `validate` method is used.
`operator-courier` should return a non zero exit code when the input
is invalid.

Fix:
Raise a ValueError in the build_and_verify method when the bundle isn't
valid.
Remove the empty bundle check in the push method, since the underlying
build_and_verify method will raise the necessary exception

Fixes https://github.com/operator-framework/operator-courier/issues/46


kevinrizza:
@awgreene please review when you get a chance

PR #48, opened 2019-03-08
Update to str.format()

awgreene:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/home/agreene/.local/share/virtualenvs/operator-courier-8e3EdCOa/bin/python"
```
kevinrizza: Let's not commit the settings.json for vscode.

awgreene: Thanks for catching that.

operatorcourier/identify.py
```diff
@@ -29,7 +29,7 @@ def get_operator_artifact_type(operatorArtifactString):
                 artifact_type = operatorArtifact["kind"]
                 artifact_name = operatorArtifact['metadata']['name']
             if artifact_type is not None:
-                logger.info('Parsed %s: %s', artifact_type, artifact_name)
```
MartinBasti: NACK, logging should stay lazy.

I mas set loglevel=ERROR, in first case logger.info() is like noop, in your case a new string is created and then dropped.

operatorcourier/nest.py
```diff
@@ -52,7 +52,7 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
         if not os.path.exists(csv_folder):
             os.makedirs(csv_folder)
 
-        csv_path = '%s/%s.clusterserviceversion.yaml' % (csv_folder, csv_name)
```
MartinBasti: I have strong preferrence to use `os.path.join()` for manipulation with filenames.
```python
csv_path = os.path.join(csv_folder, '{}.clusterserviceversion.yaml'.format(csv_name) )
```

MartinBasti: ... or `Path` from `pathlib` because this is pure py3
https://docs.python.org/3/library/pathlib.html#module-pathlib

operatorcourier/identify.py
awgreene: Sorry - I don't understand what you mean here.

.vscode/settings.json
MartinBasti: I use global gitignore for this purpose, makes my life much better.

https://stackoverflow.com/questions/7335420/global-git-ignore

PR #49, opened 2019-03-08
Verify packageName in bundle matches repository name

jeremylinlin:
Verification step added to ensure the packageName defined in package.yaml matches the quay.io repo name given as a parameter to the push command.

### DIFF
--- a/operatorcourier/api.py
+++ b/operatorcourier/api.py
@@ -20,7 +20,7 @@ logger = logging.getLogger(__name__)
 
 
 def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
-                     validation_output=None):
+                     validation_output=None, repository=None):
     """Build and verify constructs an operator bundle from
     a set of files and then verifies it for usefulness and accuracy.
 
@@ -30,6 +30,7 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
     :param yamls: List of yaml strings to create bundle with
     :param ui_validate_io: Optional flag to test operatorhub.io specific validation
     :param validation_output: Path to optional output file for validation logs
+    :param repository: Repository name for the application
     """
 
     if source_dir is not None and yamls is not None:
@@ -49,7 +50,8 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
 
     bundle = BuildCmd().build_bundle(yaml_files)
 
-    valid, validation_results_dict = ValidateCmd(ui_validate_io).validate(bundle)
+    valid, validation_results_dict = ValidateCmd(ui_validate_io).validate(bundle,
+                                                                          repository)
 
     if not valid:
         bundle = None
@@ -82,7 +84,8 @@ def build_verify_and_push(namespace, repository, revision, token,
     :param validation_output: Path to optional output file for validation logs
     """
 
-    bundle = build_and_verify(source_dir, yamls, validation_output=validation_output)
+    bundle = build_and_verify(source_dir, yamls, repository=repository,
+                              validation_output=validation_output)
 
     with TemporaryDirectory() as temp_dir:
         with open('%s/bundle.yaml' % temp_dir, 'w') as outfile:
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -45,18 +45,19 @@ class ValidateCmd():
         self.validation_json['errors'].append(message % args)
         logger.error(message, *args, **kwargs)
 
-    def validate(self, bundle):
+    def validate(self, bundle, repository=None):
         """validate takes a bundle as a dictionary and returns a boolean value that
         describes if the bundle is valid. It also logs verification information when
         there are invalid sections in the bundle.
 
         :param bundle: Dictionary of bundle value
+        :param repository: Repository name for the application
         """
         logger.info("Validating bundle.")
 
-        return self._bundle_validation(bundle), self.validation_json
+        return self._bundle_validation(bundle, repository), self.validation_json
 
-    def _bundle_validation(self, bundle):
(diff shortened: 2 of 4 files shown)


operatorcourier/api.py
```diff
@@ -19,14 +19,16 @@
...
-def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False):
+def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False, repository=None):
     """Build and verify constructs an operator bundle from
     a set of files and then verifies it for usefulness and accuracy.
 
     It returns the bundle as a string.
 
     :param source_dir: Path to local directory of yaml files to be read.
     :param yamls: List of yaml strings to create bundle with
+    :param ui_validate_io: TODO
```
kevinrizza: Let's leave documentation for this parameter for a separate pr please.

operatorcourier/validate.py
```diff
@@ -208,43 +216,48 @@ def _pkgs_validation(self, bundleData):
...
-                        try:
-                            csvName = csv["metadata"]["name"]
-                            csvNames.append(csvName)
-                        except KeyError:
-                            pass
-                    for channel in channels:
-                        if "name" not in channel:
-                            logger.error("package channel.name not defined.")
-                            valid = False
+        pkg = pkgs[0]
```
kevinrizza: What about the case where pkgs == 0 ?

jeremylinlin: @kevinrizza It should already be handled [here](https://github.com/operator-framework/operator-courier/blob/master/operatorcourier/validate.py#L254) before the pkg validator is called.

operatorcourier/api.py
jeremylinlin: Sure. Am I supposed to remove this line?

kevinrizza: that is what I mean, yes.

operatorcourier/validate.py
kevinrizza: @jeremylinlin we can't rely on other validation methods. what if they were reordered or that other method was refactored?

jeremylinlin: @kevinrizza Ah that makes sense. I'll fix it. Thanks Kevin!

operatorcourier/validate.py
```diff
@@ -232,43 +240,49 @@ def _pkgs_validation(self, bundleData):
...
 
-        for pkg in pkgs:
-            if "packageName" in pkg:
-                logger.info("Evaluating package %s", pkg["packageName"])
-            else:
-                self._log_error("packageName not defined.")
-                valid = False
+        num_pkgs = len(pkgs)
+        if num_pkgs != 1:
+            logger.error('Only 1 package is expected to exist per bundle, but got %d.', num_pkgs)
```
awgreene: This should use the _log_error method

jeremylinlin: Thanks @awgreene. Great catch! Updated the PR.

operatorcourier/validate.py
```diff
@@ -75,6 +76,14 @@ def _bundle_validation(self, bundle):
...
             validationDict[self.csvKey] &= self._type_validation(
                 bundleData, self.csvKey, self._ui_validation_io, True)
+        if validationDict[self.pkgsKey] and repository is not None:
+            packageName = bundleData['packages'][0]['packageName']
+            match = repository == packageName
+            if not match:
+                self._log_error('The packageName (%s) in bundle does not match '
+                                'repository name (%s) provided as command line argument.',
+                                packageName, repository)
+            validationDict[self.pkgsKey] &= match
```
kevinrizza: @jeremylinlin 
Why are we doing this bitwise and here? shouldn't we only define behavior in the case that they don't match?

jeremylinlin: Thanks @kevinrizza, it makes sense. I have simplified the code.

operatorcourier/validate.py
```diff
@@ -45,18 +45,19 @@ def _log_error(self, message, *args, **kwargs):
...
         logger.error(message, *args, **kwargs)
 
-    def validate(self, bundle):
+    def validate(self, bundle, repository=None):
         """validate takes a bundle as a dictionary and returns a boolean value that
         describes if the bundle is valid. It also logs verification information when
         there are invalid sections in the bundle.
 
-        :param bundle: Dictionary of bundle value
+        :param bundle:     Dictionary of bundle value
```
SamiSousa: nit: Does this indent need to match for all params?

operatorcourier/validate.py
```diff
@@ -75,6 +76,13 @@ def _bundle_validation(self, bundle):
         if self.ui_validate_io:
             validationDict[self.csvKey] &= self._type_validation(
                 bundleData, self.csvKey, self._ui_validation_io, True)
+        if validationDict[self.pkgsKey] and repository is not None:
```
SamiSousa: So this check will only run for the `push` command, correct?

kevinrizza:
> One note: I know we don't want to introduce breaking changes, but is it possible to have a CLI flag that let's us skip entering the name operator bundle repo and instead uses the packageName found in the bundle?This is a nice-to-have, @kevinrizza let me know what you think.

@SamiSousa I don't want to do this because, while we technically have a soft requirement that packageName and repository name match, that's something that could be removed later. I would prefer not to introduce a feature that we may have to remove later.

operatorcourier/validate.py
jeremylinlin: @SamiSousa It seems that we don't have this indent in other places. I'll fix it.

operatorcourier/validate.py
jeremylinlin: That's correct. Because the repository argument is only required for push command.

PR #53, opened 2019-03-12
Remove vscode settings folder

kevinrizza:


jeremylinlin:
@kevinrizza I think we should also add the `.vscode` folder to `.gitignore` so that it won't be added again in the future.

PR #57, opened 2019-03-12
Update api method docs

kevinrizza:
Problem:
Some of the api methods are missing documentation on all parameters

Solution:
Update param docs for build_and_verify()
and build_verify_and_push() to include all parameter


kevinrizza:
@jeremylinlin

PR #59, opened 2019-03-12
Update .gitignore to ignore VS Code and IntelliJ settings folder

jeremylinlin:


csomh:
Just a note: choosing an editor is up to each contributor, and not a project specific thing. 

As such local configuration (`$HOME/.gitignore`) is much more appropriate place to ignore these files and folders.

kevinrizza:
> Just a note: choosing an editor is up to each contributor, and not a project specific thing.
> 
> As such local configuration (`$HOME/.gitignore`) is much more appropriate place to ignore these files and folders.

@csomh 
Adding these to the `.gitignore` does not prevent any dev workflow from overriding them locally. I don't see harm in including these. To me, the argument that developers are free to chose their IDE is exactly the reason why these folders should be included in the `.gitignore` file. We want to prevent them from getting committed accidentally.

csomh:
@kevinrizza, this would mean that every git repo should have a copy of a long ignore list to consider all possible editors used. I would consider this a waste of storage :)

Git has a layered configuration exactly to work around this:
- if a developer uses VSCode, they probably want to ignore .vscode all the time, so they can place this rule in `~/.gitignore`.
- if a project uses `tox` for testing, it's desirable to exclude `.tox` in the project `.gitignore`, as all clones of the repo should ignore that.

Or maybe `man gitignore` explains this better:

> Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

kevinrizza:
@csomh I understand the use and purpose of a layered `.gitignore` file in general, I do not believe that is the source of this disconnect. It's not that I don't understand your point, it's just that I disagree with it. My personal opinion is that a repository owner should have the discretion to purposefully exclude certain folders and types, and using tooling in git (like the `.gitignore` file) is the exact purpose of those tools.

You are correct that if the purpose of this pr was to "Exclude all possible editor files", that we would be failing to achieve that. I would argue that that is an impossible goal, since on any given day the list of code editor is `n+1`.  But my argument is that the purpose of this pr is not to exclude all possible editor files, it is specifically to exclude the common editor files that the frequent contributors to this repository currently use. Specifically, to avoid accidentally checking in files like what occurred here: https://github.com/operator-framework/operator-courier/pull/27

That pr was reviewed by three different contributors, but it still managed to get one of these files in. That proves that manual intervention was not enough, and a tool like this would have prevented this from happening. In this particular case (which is not only possible but also entirely likely to happen again) I think it is not a waste of space or processing time.

csomh:
I understand the practicality of it, @kevinrizza. Nevertheless, I consider allowing such a change a missed opportunity to help devs master the tools they are working with.

kevinrizza:
@csomh I can appreciate the desire to educate. But `.gitignore` is ultimately a tool to make repository maintenance easier. Unfortunately, if we made all of our decisions in our repositories toward education it would slow down our ability to maintain by a significant margin.

PR #61, opened 2019-03-12
Add tox installation as prereq in contributing doc

jeremylinlin:


docs/contributing.md
```diff
@@ -6,13 +6,24 @@ The following is a set of best practices for contributing to the operator-courie
 This project uses [pipenv](https://github.com/pypa/pipenv) to manage dependencies and create virtual environments for development environment.
 
 ### Prerequisites
-Python3
```
kevinrizza: Why add the space here? It just adds noise to the pr.

docs/contributing.md
```diff
@@ -6,13 +6,24 @@ The following is a set of best practices for contributing to the operator-courie
...
+
+- tox
+
+  You could either install it globally
+  
+  - `sudo dnf install tox` 
+  
+  or install inside your pip environment (you will have to install it every time you enter the virtual environment)
+  
+  - `pip install tox`
```
kevinrizza: Maybe we should just give one of these? There are a ton of ways to install things in lots of different environments (`apt-get`, `brew`, `yum` etc.). Can we just list the Fedora `dnf` option for simplification here? At least that is in line with the rest of the prereqs here.

jeremylinlin: @kevinrizza Thanks Kevin. Simplified and updated.

PR #62, opened 2019-03-12
Update README about minimum supported version of Python

jeremylinlin:
/cc @kevinrizza @awgreene

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 The Operator Courier is used to build, validate and push Operator Artifacts.
 
+Operator Courier is currently supported on Python 3.6 and above.
+
 ## Usage
 
 ### Command Line Interface
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,7 @@ setup(
     'regex',
     'validators'
   ],
+  python_requires='>=3.6, <4',
   setup_requires=['pytest-runner'],
   tests_require=tests_require,
   extras_require={
@@ -36,4 +37,8 @@ setup(
   long_description_content_type="text/markdown",
   license="Apache License 2.0",
   include_package_data=True,
+  classifiers=[
+    'Programming Language :: Python :: 3.6',
+    'Programming Language :: Python :: 3.7',
+  ],
 )


MartinBasti:
Maybe you can also use PyPI classifiers to make it exact in PyPI.

https://github.com/release-engineering/operators-manifests-push-service/blob/master/setup.py#L27

https://pypi.org/classifiers/

MartinBasti:
And `python_requires='>=3.6, <4'` in setup.py

jeremylinlin:
Hi @MartinBasti, thank you for referring me to PyPI classifiers. Regarding setting up `python_requires`, does this prevent our users from using the Python version below 3.6? Although we explicitly say we support Python 3.6+, tox tests actually pass on Python 3.0+, meaning courier may or may not run on versions below 3.6 as well. Do we need to restrict our users' choice of versions?

MartinBasti:
@jeremylinlin https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

```
If your project only runs on certain Python versions, setting the python_requires argument to the appropriate PEP 440 version specifier string will prevent pip from installing the project on other Python versions
```

jeremylinlin:
@MartinBasti Yeah I think my point is we probably should not prevent our users from using Python 3.6- by setting `python_requires`. We don't say we support Python 3.6- because we don't want to advertise it since we don't test on these versions.

MartinBasti:
We are protecting users from shooting their legs with untested code. 

We can:
* wait for backport request
* add testing for more py3 versions

I prefer to limit support for 3.6+ and not give up all new cool python features (I'd not go with 3.3 and lower ever)

jeremylinlin:
Hey @MartinBasti, thank you very much for your comments. It makes sense and I have updated the PR accordingly.

README.md
```diff
@@ -2,6 +2,8 @@
 
 The Operator Courier is used to build, validate and push Operator Artifacts.
 
+We currently support Python 3.6 and above.
```
kevinrizza: Instead of using `We` here, can we use `Operator-Courier` as the subject of the sentence?

`Operator Courier is currently supported on Python 3.6 and above`

jeremylinlin: Thanks @kevinrizza. Updated.

PR #63, opened 2019-03-12
Clean up additional folders generated by tox

jeremylinlin:
/cc @kevinrizza @awgreene

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,3 +8,5 @@ clean:
 	rm -rf operator_courier.egg-info
 	rm -rf build
 	rm -rf dist
+	rm -rf .eggs
+	rm -rf .tox


csomh:
Based on the docs, `make` is used for builds and deployments. `tox` is used for texting? Why does `make clean` need to touch `tox` artifacts?

kevinrizza:
@csomh Running tests is a necessary part of determining if a build should run successfully. I do not see a problem with using `make` to clean up any artifacts left over from running any temporary workflow that is associated with building or releasing from a repository. `make clean` is a common pattern in dev workflows, as is `make unit` or `make test`

csomh:
@kevinrizza, ok, I see now. Thanks!

PR #64, opened 2019-03-13
The `push-to-quay` script support running on MACos

jianzhangbjz:
Support it for MAC, fixed below errors on MAC:
```console
base64: invalid option -- w
Usage:	base64 [-hvD] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -D, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
```

### DIFF
--- a/scripts/push-to-quay
+++ b/scripts/push-to-quay
@@ -21,7 +21,11 @@ trap cleanup EXIT
 
 tar czf ${FILENAME}.tar.gz ${FILENAME}.yaml
 
-BLOB=$(cat ${FILENAME}.tar.gz | base64 -w 0)
+if [ $(uname) == "Darwin" ];then
+    BLOB=$(cat ${FILENAME}.tar.gz | base64 -b 0)   
+else
+    BLOB=$(cat ${FILENAME}.tar.gz | base64 -w 0)
+fi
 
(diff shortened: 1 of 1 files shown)


jianzhangbjz:
> Furthermore, the title of this PR is inaccurate.

Sorry, my mistake. Changed it.

> If operator-courier is going to be supported on macos, the correct method of doing that is to test the python package and ensure that it runs successfully on that platform. 

I used the `operator-courier` to `verify`/`push` the bundles to the Quay, it worked well on MACos.

>  The only way I would accept this pr is if we added macos to our .travis.yml file and found that either the tests passed, or also made changes to the source if the tests failed.

OK, I saw #70 for this. If any further testing on Mac os, just let me know.

kevinrizza:
@jianzhangbjz

No problem, thanks for pointing out the limitation in our support! :)

SamiSousa:
@kevinrizza @jianzhangbjz Should this PR be closed since it modifies a deprecated/historical script?

kevinrizza:
@SamiSousa Good point. Closing this pr as `no-op`.

PR #65, opened 2019-03-13
Fix: UI validation functions

SamiSousa:
- Fix name of 'Deep Insights' capability
- Set key to 'capabilities' for error logging
- Add regex for alternate version format
- Change validation flag to use dashes instead of underscores

The first two issues I encountered when running the courier against
community-operators PRs.
The regex is important for versions that use a string after the
semver.
I also changed the validation CLI flag to use dashes as is the
convention for CLI flags. If using underscores was intended I can remove this change.

### DIFF
--- a/operatorcourier/cli.py
+++ b/operatorcourier/cli.py
@@ -59,7 +59,7 @@ These are the commands you can use:
         parser.add_argument('source_dir', help='Path of your directory of yaml '
                             'files to bundle. Either set this or '
                             'use the files argument for bundle data.')
-        parser.add_argument('--ui_validate_io',
+        parser.add_argument('--ui-validate-io',
                             help='Validate bundle for operatorhub.io UI',
                             action='store_true')
         parser.add_argument('--validation-output', dest='validation_output',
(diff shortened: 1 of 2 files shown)


operatorcourier/cli.py
```diff
@@ -59,7 +59,7 @@ def verify(self):
         parser.add_argument('source_dir', help='Path of your directory of yaml '
                             'files to bundle. Either set this or '
                             'use the files argument for bundle data.')
-        parser.add_argument('--ui_validate_io',
+        parser.add_argument('--ui-validate-io',
```
kevinrizza: This should be a separate pr, since we are introducing a breaking change on a released feature.

operatorcourier/validate.py
```diff
@@ -467,7 +468,7 @@ def get_alm_kinds(alm_examples):
         # capabilities check
         if not is_capability_level(annotations["capabilities"]):
             self._log_error("metadata.annotations.capabilities %s is not a "
-                            "valid capabilities level", annotations["capability"])
+                            "valid capabilities level", annotations["capabilities"])
             valid = False
 
         return valid
```
kevinrizza: The three changes in this file all seem unrelated. Can you please submit them as separate pull requests?

SamiSousa: Sure thing!

SamiSousa:
Following @kevinrizza 's suggestion and closing in favor of separate PRs for each change:
#67 for the capabilities change
#68 for the regex update
#69 for the CLI flag update

PR #66, opened 2019-03-13
Error reporting

chmeliik:


operatorcourier/errors.py
```diff
@@ -0,0 +1,46 @@
...
+class OpCourierBadArtifact(OpCourierError):
+    """File is a valid yaml, but not a valid CSV, CRD or Package"""
+    pass
+
+
+class OpCourierBadBundle(OpCourierError):
+    """Invalid bundle (e.g. missing CSV/CRD/Package).
+    Contains the info collected during validation of the bundle.
+    """
+    def __init__(self, msg, validation_dict):
```
MartinBasti: I'd call this `validation_results`, we shouldn't name variables with types in python

operatorcourier/errors.py
```diff
@@ -0,0 +1,46 @@
...
+        super().__init__(msg)
+        self.validation_dict = validation_dict
+
+
+class OpCourierQuayError(OpCourierError):
+    """An error occurred while pushing bundle to Quay.io"""
+    def __init__(self, msg, code=500):
+        """
+        :param msg: the message of this exception
+        :param code: http status code returned by request
```
MartinBasti: s/request/response/ maybe ???

tests/test_api.py
```diff
@@ -37,3 +38,29 @@ def test_make_bundle_with_yaml_list(yaml_files, expected):
...
+    ["tests/test_files/bundles/api/bundle1/crd.yml",
+     "tests/test_files/bundles/api/bundle1/csv.yaml"]
+])
+def test_make_bundle_invalid(yaml_files):
+    yamls = []
+    for file in yaml_files:
+        with open(file, "r") as yaml_file:
+            yamls.append(yaml_file.read())
+
+    with pytest.raises(OpCourierBadBundle) as e_info:
```
MartinBasti: why `e_info` and not just `e` or `exc`?

operatorcourier/errors.py
```diff
@@ -0,0 +1,46 @@
+"""
+operatorcourier.errors
+
+Defines custom operator-courier exceptions.
+"""
+
+
+class OpCourierError(ValueError):
```
MartinBasti: Could we add unit test that `OpCourierError` (and children) are subclass of `ValueError` to make sure we will keep this BW compatibility. Unless @kevinrizza agree to break compatibility

MartinBasti:
@chmeliik Good job, I had just minor comments! Let's wait for owners if they agree with such changes :-)

MartinBasti:
Maybe public API methods should also have documented which exceptions they raise.

tests/test_api.py
chmeliik: `pytest.raises()` wraps the actual exception in an `ExceptionInfo` object, hence `e_info`. I'm down to change it though if that makes it more readable (tomorrow, that is :smile: )

operatorcourier/errors.py
kevinrizza: Maintaining backwards compatibility is a major goal in the short term, since we have only recently released. Making breaking changes should be reserved for a specific release goal, which this pr is not aiming for. Please maintain backwards compatibility.

operatorcourier/errors.py
```diff
@@ -0,0 +1,46 @@
...
+        :param msg: the message of this exception
+        :param validation_dict: info collected during validation of bundle
+        """
+        super().__init__(msg)
+        self.validation_dict = validation_dict
+
+
+class OpCourierQuayError(OpCourierError):
+    """An error occurred while pushing bundle to Quay.io"""
+    def __init__(self, msg, code=500):
```
kevinrizza: Why do we have a default here? If there is an error returned from the server, we should always have an error code (even if the error is that the server is not available -- which is not 500).

tests/test_api.py
MartinBasti: Ah I didn't realize that, it is ok then :D

operatorcourier/errors.py
chmeliik: That's fair. Will remove default value.

MartinBasti: I was thinking about this, I'd really like to get content from quay what is wrong with error code and quay error JSON if possible.

We may actually split this into more exceptions:
* OpCourierQuayError -- generic quay error
* OpCourierQuayCommunicationError(OpCourierQuayError) - random error that happened during communication
*OpCourierQuayErrorResponse(OpCourierQuayError) -- error returned from quay with quay details, basically catching HTTPException from requests.

chmeliik:
Updated PR, cc @kevinrizza  @MartinBasti  @awgreene

kevinrizza:
@chmeliik Can you please squash your commits?

chmeliik:
@kevinrizza Can do, do you want all the changes in a single commit?

kevinrizza:
> @kevinrizza Can do, do you want all the changes in a single commit?

@chmeliik Yes please.

tests/test_errors.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@pytest.mark.parametrize('op_courier_exception', [
+    OpCourierBadBundle,
+    OpCourierBadYaml,
+    OpCourierBadArtifact,
+    OpCourierQuayError,
+    OpCourierError
+])
+def test_exception_backwards_compatibility(op_courier_exception):
```
MartinBasti: Please add docstring that we are testing BW compatiblity

operatorcourier/errors.py
MartinBasti: @chmeliik This got +1 from Kevin

tests/test_errors.py
chmeliik: Done

operatorcourier/errors.py
chmeliik: Understood, will get on it tomorrow

chmeliik:
Added specific Quay errors, cc @kevinrizza @MartinBasti @awgreene

operatorcourier/errors.py
```diff
@@ -0,0 +1,63 @@
...
+        :param msg: the message of this exception
+        :param validation_info: info collected during validation of bundle
+        """
+        super().__init__(msg)
+        self.validation_info = validation_info
+
+
+class OpCourierQuayError(OpCourierError):
+    """An error occurred while pushing bundle to Quay.io"""
+    def __init__(self, msg, code):
```
MartinBasti: Code is not supposed to be here, connection error has no HTTP code for example.

operatorcourier/push.py
```diff
@@ -42,15 +47,33 @@ def _push_to_registry(self, namespace, repository, release, bundle, auth_token):
...
-
-            raise ValueError("Failed to push to app registry: {}".format(msg))
+
+        if r.status_code == 200:
+            return
+
+        logger.error(r.text)
+        msg = "Failed to push to app registry: {}"
+
+        try:
```
MartinBasti: You may not even get here, IIRC Connection errors are raise directly by `requests.post()`, only HTTP status from reponse is raised by `raise_for_status`

chmeliik: Would it then be safe to assume that if `requests.post()` doesn't fail, the response will always contain valid JSON content from  Quay? Or should I keep using `OpCourierQuayCommunicationError` for that and use `OpCourierQuayError` for exceptions raised by `requests.post()`?

kevinrizza: The API unfortunately does not have a public spec and the source code is not yet open. However, I believe it is a fair assumption that, if requests.post() on that endpoint returns some kind of success that there will always be a json response, even if that response is empty. I haven't seen any behavior otherwise.

MartinBasti: It needs something like this:
```python
try:
    r = request.post(.....)
except RequestsException as e:  # or Exception
   raise OpCourierQuayCommunicationError(str(e))
else:
   if r.status_code != 200:
     try:
        r_json = r.json()
     except Exception:
        r_json = {}
     raise OpCourierQuayErrorResponse('something', code, r_json)
```

chmeliik: That looks reasonable. The only potential problem could that raising `OpCourierQuayCommunicationError` instead of `RequestException` would break backward compatibility. Of course, `OpCourierQuayError` could inherit from `RequestException` as well, though that would still break code like this: 
```python
try:
    api.build_verify_and_push()
except ValueError:
    # OpCourierQuayError gets caught here because it inherits from both
except RequestException:
    # and this doesn't happen
```
I'm probably making things unnecessarily complicated, but I'd still like to get your thoughts on it @MartinBasti  @kevinrizza

kevinrizza: @chmeliik @MartinBasti Personally I think anything where `r.status_code != 200` should except one error type that inherits from `ValueError` and anything where `requests.post()` throws an exception itself should except another error type that inherits `RequestException`. To me I thought that was the whole point of separating these error types.

MartinBasti: Ok, it could be done in this way

chmeliik:
Updated PR. Had to change the exception hierarchy a bit to allow for an exception that would inherit from the base exception but not from `ValueError`. Added a couple simple tests to make sure I didn't mess it up.
cc @MartinBasti @kevinrizza

chmeliik:
Rebased on master

PR #68, opened 2019-03-13
Fix: Expand regular expression for version

SamiSousa:
- Combined regex patterns
- Added optional version suffix

This adds support for versions that use a string after the semver.
Replaces #65

### DIFF
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -1,6 +1,6 @@
 import logging
 import json
-import re
+import semver
 
 import validators as v
 
@@ -386,9 +386,11 @@ class ValidateCmd():
             return True
 
         def is_version(field):
-            pattern1 = re.compile(r'v(\d+\.)(\d+\.)(\d)')
-            pattern2 = re.compile(r'(\d+\.)(\d+\.)(\d)')
-            return pattern1.match(field) or pattern2.match(field)
+            try:
+                semver.parse(field)
+            except ValueError:
+                return False
+            return True
 
         def is_capability_level(field):
             levels = [
@@ -494,8 +496,8 @@ class ValidateCmd():
 
         # version check
         if not is_version(spec["version"]):
-            self._log_error("spec.version %s is not a valid version "
-                            "(example of a valid version is: v1.0.12)",
+            self._log_error("spec.version %s is not a valid semver "
+                            "(example of a valid semver is: 1.0.12)",
                             spec["version"])
             valid = False
 
--- a/tests/test_validate.py
+++ b/tests/test_validate.py
@@ -50,8 +50,8 @@ def test_ui_valid_bundle_io(bundle, expected_validation_results_dict):
         ("tests/test_files/bundles/verification/ui.invalid.bundle.yaml",
             {'errors': [
                 "csv.spec.links must be a list of name & url pairs.",
-                "spec.version invalid is not a valid version "
-                "(example of a valid version is: v1.0.12)",
+                "spec.version invalid is not a valid semver "
+                "(example of a valid semver is: 1.0.12)",
                 "metadata.annotations.capabilities invalid "
                 "is not a valid capabilities level",
                 "spec.icon[0].mediatype image/invalid is not "
--- a/Pipfile
+++ b/Pipfile
@@ -4,12 +4,11 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-
 validators = "*"
 requests = "*"
 pytest = "*"
 pyyaml = "*"
 twine = "*"
-
+semver = "*"
 
 [dev-packages]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "18523475689e3183d2da8dfea9f6f6e55520811e8a2812d7415feb4738a08cc9"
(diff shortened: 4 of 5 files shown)


kevinrizza:
@SamiSousa Could you add a unit test that ensures success for the version suffix condition?

operatorcourier/validate.py
```diff
@@ -371,9 +371,8 @@ def is_email(field):
             return True
 
         def is_version(field):
-            pattern1 = re.compile(r'v(\d+\.)(\d+\.)(\d)')
-            pattern2 = re.compile(r'(\d+\.)(\d+\.)(\d)')
-            return pattern1.match(field) or pattern2.match(field)
+            pattern = re.compile(r'(v)?(\d+\.)(\d+\.)(\d)(-[a-z0-9\.])*')
```
MartinBasti: Do you need all `()`? You are not returning any substring from field

```python
r'v?\d+[.]\d+[.]\d(-([a-z0-9][.]?)+)*'
```
I miss unittests.
Actually this is not working as you expect. `(-[a-z0-9\.])*` matches `-a-.-2` instead `-a.2`. How about uppercase letters and underscores, are they allowed?
Also If I remember correctly quay doesn't allow leading zeroes

SamiSousa: For the AppRegistry, quay allows any valid semver. Things like `0.0.1` and `1.21.5-beta1` are valid.

SamiSousa: I agree that we can remove extraneous groups. I'd actually prefer using a proven semver library rather than reinventing the regex.

SamiSousa: Scratch what I said, this version in the CSV has no relation to quay's AppRegistry, my mistake.
In any case, there should be a doc that outlines what format this version field can take. From experience I have seen versions that use a semver followed by a tag. I agree that `-a-.-2` doesn't look right, so I'll update it to avoid something like that

operatorcourier/validate.py
```diff
@@ -371,9 +371,9 @@ def is_email(field):
             return True
 
         def is_version(field):
-            pattern1 = re.compile(r'v(\d+\.)(\d+\.)(\d)')
-            pattern2 = re.compile(r'(\d+\.)(\d+\.)(\d)')
-            return pattern1.match(field) or pattern2.match(field)
+            version = re.compile(r'v?\d+\.\d+\.\d+'
+                r'(-[A-Za-z0-9]+([-\.]?[A-Za-z0-9]+)*)?'
```
SamiSousa: @alecmerdler Is there a standard for what a valid CSV version field is?

alecmerdler: Just standard SemVer (`major.minor.patch`).

SamiSousa: I've seen versions such as `0.0.1` and `1.22.0-beta1`, so I've set this regex to match those as well.

kevinrizza: But `major.minor.patch` isn't the full standard for Semantic Versioning:

https://semver.org/#spec-item-9

Is prerelease tagging supported? If so, we should check for it.

operatorcourier/validate.py
SamiSousa: It'd be better to use an external semver validator such as py package [semver](https://pypi.org/project/semver/) to avoid having to test the version regex we've implemented, especially as it becomes increasingly more complex.

operatorcourier/validate.py
SamiSousa: @kevinrizza Based on what I've seen, prerelease tagging is supported, and should be checked fo. I suggest that we use an external semver validator such as py package [semver](https://pypi.org/project/semver/)

kevinrizza: +1

SamiSousa:
@kevinrizza Added the `semver` package for semver parsing, updated the Pipfile appropriately. Updated the example version to match a valid version value

operatorcourier/validate.py
```diff
@@ -478,7 +480,7 @@ def is_mediatype(mediatype):
         # version check
         if not is_version(spec["version"]):
             self._log_error("spec.version %s is not a valid version "
-                            "(example of a valid version is: v1.0.12)",
+                            "(example of a valid version is: 1.0.12)",
```
kevinrizza: Can we specify something about expecting semver here?

tests/test_validate.py
```diff
@@ -46,7 +46,7 @@ def test_ui_valid_bundle_io(bundle, expected_validation_results_dict):
             {'errors': [
                 "csv.spec.links must be a list of name & url pairs.",
                 "spec.version invalid is not a valid version "
-                "(example of a valid version is: v1.0.12)",
+                "(example of a valid version is: 1.0.12)",
```
kevinrizza: Can we make sure to have both a positive and a negative test?

SamiSousa: I'll use `v<anything>` as the negative test for this field

SamiSousa: Actually, since this is already the negative, do you mean we need a positive test case?

SamiSousa: For the positive test case, isn't the `valid.bundle.yaml` sufficient? Or should I test additional valid semvers? I don't think I need to test this since we're using an external package.

SamiSousa:
@MartinBasti @kevinrizza I've rebased my PR, please let me know if additional changes are needed

PR #70, opened 2019-03-13
Update Travis CI config to test courier on macOS

jeremylinlin:
Build operator-courier on macOS with Travis CI.

Because testing Python projects on macOS is not natively supported by Travis CI, I use a [workaround](https://pythonhosted.org/CodeChat/.travis.yml.html) to make it happen. Finishing tests on macOS with Travis CI will take around 10min, due to the time-consuming `brew update` and `brew install` operations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,36 @@ matrix:
       python: "3.7"
       env: TOXENV=py37
       after_success: coveralls
+    - os: osx
+      language: generic
+      env:
+        - PYTHON=3.6.0
+        - TOXENV=py36
+    - os: osx
+      language: generic
+      env:
+        - PYTHON=3.7.0
+        - TOXENV=py37
     - dist: xenial
       python: "3.7"
       env: TOXENV=flake8
+
+# Installing Python on macOS for Travis CI: https://pythonhosted.org/CodeChat/.travis.yml.html
+before_install: |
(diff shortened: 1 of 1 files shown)


MartinBasti:
What are we trying here? I haven't noticed any platform dependent code in operator-courier

jeremylinlin:
@MartinBasti See https://github.com/operator-framework/operator-courier/pull/64. I'm updating the travis ci config to run on macOS to make sure it works on that platform as well.

MartinBasti:
Do we use any platform specific python modules? Will we test windows too?

jeremylinlin:
@MartinBasti @kevinrizza
> Do we use any platform specific python modules?

Maybe not for now.

> Will we test windows too?

Based on our discussion today, we will support courier on macOS but not Windows.

csomh:
This will make Travis builds to be 5 to 10 times slower (1-2 minutes vs 10+ minutes), and as such delay feedback for devs. 

As @MartinBasti pointed out, there is no platform specific code in `operator-courier`, so I'm not sure about the value of running the tests on Mac.

kevinrizza:
@csomh @MartinBasti You are right, today we do not have any platform specific code. But that doesn't guarantee that we won't have some committed in the future, especially given that the CLI does filesystem modification (take a look at the `nest` command as an example).

Furthermore, 10+ minutes is not an enormous change, given that code review should take more than an order of magnitude longer than that.

PR #72, opened 2019-03-18
UI validation - adding icon mediatype check

galletti94:
check that mediatype is valid


operatorcourier/validate.py
```diff
@@ -390,6 +390,9 @@ def get_alm_kinds(alm_examples):
             for example in alm_examples:
                 res.append(example["kind"])
             return res
+        
+        def is_mediatype(mediatype):
+            return mediatype in ["image/gif", "image/jpeg", "image/png", "image/svg+xml"]
```
SamiSousa: Is this an exhaustive list of our supported image types?

operatorcourier/validate.py
```diff
@@ -469,5 +472,28 @@ def get_alm_kinds(alm_examples):
...
                             "valid capabilities level", annotations["capability"])
             valid = False
+        
+        # icon check
+        if "icon" in spec:
+            icon = spec["icon"]
+            if isinstance(icon, (list,)):
+                if len(icon) == 1:
+                    if len(icon[0].keys()) == 2:
+                        if "base64data" in icon[0] and "mediatype" in icon[0]:
```
SamiSousa: Will we also check that the base64data is valid?

operatorcourier/validate.py
galletti94: as far as I know yes

operatorcourier/validate.py
galletti94: I don't think that's really possible. I can check that the starting bits correctly match what the mediatype is defined to be but as far as I know there is no way to check that the image is "valid". I suggest just going to operatorhub.io/preview for a visual confirmation

PR #73, opened 2019-03-18
Add basic integration tests with behave

jeremylinlin:
The integration tests currently only test the `verify` command. For the `push` command, we need to set up quay access token as environment variable. Due to some [restrictions](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings) from Travis CI,

> Encrypted environment variables are not available to pull requests from forks due to the security risk of exposing such information to unknown code.

> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

We decided to include it later and run those tests directly on master, and trigger Travis CI builds periodically with cron jobs. Details are yet to be decided.

### DIFF
--- a/features/environment.py
+++ b/features/environment.py
@@ -1,0 +1,21 @@
+from subprocess import check_call
+
+
+def before_tag(context, tag):
+    if tag == 'cli':
+        commands = [
+            'make release',
+            'pip install dist/operator_courier-*-none-any.whl',
+            'operator-courier -v',
+        ]
+
+        for command in commands:
+            exit_code = check_call(command, shell=True)
+            assert exit_code == 0
+
+
+def after_tag(context, tag):
+    if tag == 'cli':
+        cmd = 'make clean'
+        exit_code = check_call(cmd, shell=True)
+        assert exit_code == 0
--- a/Pipfile
+++ b/Pipfile
@@ -10,6 +10,7 @@ requests = "*"
 pytest = "*"
 pyyaml = "*"
 twine = "*"
+behave = "*"
 
 
 [dev-packages]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "18523475689e3183d2da8dfea9f6f6e55520811e8a2812d7415feb4738a08cc9"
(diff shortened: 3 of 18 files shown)


features/steps/push.py
```diff
@@ -0,0 +1,35 @@
+from behave import step, use_step_matcher
+import subprocess as sp
+import uuid
+import requests
+
+use_step_matcher("re")
+
+# TODO: Quay API to remove particular application?
```
SamiSousa: This may be a question for forum-quay, unless @kevinrizza has already asked about this.

jeremylinlin: Hey @SamiSousa, thanks for letting me know this. This is still WIP code so I will make sure this is removed when it's finalized.

csomh:
Ok, I'll be the devil's advocate, again :)

I strongly discourage using behave for the purpose of integration testing, as it's not really suitable to test these kind of applications on the long run.

I had a really bad experience setting up an entire integration test suite for [OSBS](https://osbs.readthedocs.io/en/latest/), which was based on behave, and which was ultimately migrated to pytest in order to improve maintainability and ease further development.

Behave introduces a layer of abstraction which is hard to work around when testing APIs or CLIs. Behave is suitable for testing applications or services, where the functionality can be described in relatively simple _human_ terms. Operator Artifacts are beyond that, I think, and as such the language of the scenarios will become convoluted and hard to follow.

jeremylinlin:
Hi @csomh, thank you very much for your advice. Could you provide some concrete examples or scenarios why `behave` might be poorly suited for operator-courier? I current don't have issues with this framework as it does what I want it to do, and switching to something else would take extra effort.

Also, I have set up a group chat on slack with @kevinrizza, feel free to reach out to us there to discuss in depth.

features/environment.py
```diff
@@ -0,0 +1,21 @@
+from subprocess import check_call
+
+
+def before_tag(context, tag):
+    if tag == 'cli':
+        commands = [
```
csomh: IIRC, `tox` already implicitly takes care of building and installing the tool in the environment, unless it's configured not to do it. So I think, this won't be required.

jeremylinlin: Thanks @csomh, I'll keep that in mind for my next push.

jeremylinlin:
Closing it because we decide to use `pytest` instead of `behave` for integration tests.

PR #75, opened 2019-03-18
Bug fix capability is not a valid field

galletti94:
capability should be capabilities


SamiSousa:
This addressed in #67

kevinrizza:
Good point, there is already a PR open for this. Closing as dup.

PR #78, opened 2019-03-20
Adding categories check

galletti94:
Adding a check to verify that the categories listed are valid categories


tests/test_files/bundles/verification/ui.invalid.bundle.yaml
```diff
@@ -39,12 +39,12 @@ data:
                 }
               
             ]
-          categories: openshift required
```
kevinrizza: Won't these already fail?

galletti94: Yes, just wanted to be consistent with the invalid values

PR #79, opened 2019-03-20
Add a class that would allow access to results from verification.

chmeliik:
[OMPS](https://github.com/release-engineering/operators-manifests-push-service) would like access to results from bundle verification.

The idea is that you would use the new `build_bundle()` api function to get an `OperatorBundle` object. This object would allow you to `validate()` and `push()` and would store the info from validation.

The existing api functions were rewritten to demonstrate.

### DIFF
--- a/operatorcourier/api.py
+++ b/operatorcourier/api.py
@@ -8,11 +8,8 @@ when using the operator-courier package
 import os
 import logging
 from tempfile import TemporaryDirectory
-import yaml
-import json
-from operatorcourier.build import BuildCmd
-from operatorcourier.validate import ValidateCmd
-from operatorcourier.push import PushCmd
+
+from operatorcourier.bundle import OperatorBundle
 from operatorcourier.format import format_bundle
 from operatorcourier.nest import nest_bundles
 from operatorcourier.errors import OpCourierBadBundle
@@ -39,39 +36,21 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
     :raises OpCourierBadArtifact: When a file is not any of {CSV, CRD, Package}
     :raises OpCourierBadBundle: When the resulting bundle fails validation
     """
-
-    if source_dir is not None and yamls is not None:
-        logger.error("Both source_dir and yamls cannot be defined.")
-        raise TypeError(
-            "Both source_dir and yamls cannot be specified on function call.")
-
-    yaml_files = []
-
-    if source_dir is not None:
-        for filename in os.listdir(source_dir):
-            if filename.endswith(".yaml") or filename.endswith(".yml"):
-                with open(source_dir + "/" + filename) as f:
-                    yaml_files.append(f.read())
-    elif yamls is not None:
-        yaml_files = yamls
-
-    bundle = BuildCmd().build_bundle(yaml_files)
-
-    valid, validation_results_dict = ValidateCmd(ui_validate_io).validate(bundle,
-                                                                          repository)
+    bundle = build_bundle(source_dir, yamls,
+                          repository=repository,
+                          ui_validate_io=ui_validate_io)
 
     if validation_output is not None:
-        with open(validation_output, 'w') as f:
-            f.write(json.dumps(validation_results_dict) + "\n")
+        bundle.write_validation_info(validation_output)
 
-    if valid:
-        bundle = format_bundle(bundle)
-        return bundle
+    if bundle.valid:
+        bundle_data = format_bundle(bundle.data_copy())
+        return bundle_data
     else:
         logger.error("Bundle failed validation.")
         raise OpCourierBadBundle(
             "Resulting bundle is invalid, input yaml is improperly defined.",
-            validation_info=validation_results_dict
+            validation_info=bundle.validation_info
         )
 
 
@@ -100,18 +79,13 @@ def build_verify_and_push(namespace, repository, revision, token,
 
     :raises OpCourierQuayCommunicationError: When communication with Quay fails
     :raises OpCourierQuayErrorResponse: When Quay responds with an error
-    :raises OpCourierQuayError: When the request fails in an unexpected way
     """
+    bundle = build_bundle(source_dir, yamls, repository=repository)
 
-    bundle = build_and_verify(source_dir, yamls, repository=repository,
-                              validation_output=validation_output)
+    if validation_output is not None:
+        bundle.write_validation_info(validation_output)
 
-    with TemporaryDirectory() as temp_dir:
-        with open('%s/bundle.yaml' % temp_dir, 'w') as outfile:
-            yaml.dump(bundle, outfile, default_flow_style=False)
-            outfile.flush()
-
-        PushCmd().push(temp_dir, namespace, repository, revision, token)
+    bundle.push(namespace, repository, revision, token)
 
 
 def nest(source_dir, registry_dir):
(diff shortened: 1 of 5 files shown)


kevinrizza:
@chmeliik I have an issue with the structure. Right now, the API is written in such a way that calling a single method does all the work. With this change, you would need to call one method to build the bundle, another to validate, and a third to push. What is the value gain here? Simply to inspect the bundle before running these successive commands?

We could achieve that same result by having the build_and_validate() method also return the bundle without breaking the structure of the API. Or even just implementing an additional method to build the bundle separately. I would prefer not to decouple these, they were coupled for a reason.

chmeliik:
@kevinrizza The main reason why I decided to go for this (admittedly not spectacular) solution was to leave the existing API intact. What we need is straightforward access to the information you get when validating the bundle, and returning that from`build_and_verify()` would, to my understanding, break the current API.

Another solution would be to make a method that does exactly the same thing as `build_and_verify()`, but returns the validation info as well (and to also return the validation info from `build_verify_and_push()`). I don't know if that's a better solution, it very well could be.

Or, if you are not entirely opposed to the idea of the new class, I could either:
a) Have the bundle always get validated after building, so you would only need to use `push()` afterwards (if you were even going to do that in the first place).
b) Instead of `push()`, have a `validate_and_push()` method, though that would not let you decide what to do with an invalid bundle.

Oh, and just to clarify, I'm not suggesting the `build_bundle()` method replace the current API. It would just be there in case the existing methods do not provide enough control.

MartinBasti:
@kevinrizza I really like OOP model Adam did. It gives you more power to future enhancements and it increases readability.
I agree with coupling build+verification (@chmeliik please make verification as mandratory part of build in the bundle class), but I think the push part shouldn't be coupled as there is more use cases (pushing to quay, save bundle to dir, possibly extract metadata) and each case can be method of bundle class.

Kevin would you accept when `build_and_verify()` returns that `bundle` object?

chmeliik:
Made verification part of the build process @MartinBasti

kevinrizza:
@MartinBasti @chmeliik What is the actual use case here though?

I would strongly prefer to defer this because it will require significantly more maintenance, especially as the format of the bundle changes as part of OpenShift 4.2:

https://github.com/operator-framework/operator-registry#manifest-format

The `operator-courier` right now is intended to simply be a tool that can verify and push these bundles at a high level. It isn't intended to be the source of truth for all schema, and adding all of these features for convenience will over time make maintaining multiple versions of these bundles extremely difficult. To me it is much better for compatibility if the bundle is not returned. Even if we were to go to such a model, I would prefer that the `bundle` class itself to be viewed as `__private` implementation details. If we wanted to return "something", we could set a flag that rather than use a temporary directory we could make the file permanent.

Is there a use case here besides getting data about the bundle? And if so, is there a reason that couldn't be more than simply parsing a file that is written to disk which essentially could just be parsed as a map?

MartinBasti:
@kevinrizza use case is to get validation errors from successful  validation/push.

Ok, in case there will be a new API/bundle versions we can drop this. (However I cannot imagine how to handle compatibility better than with class per bundle format version)

Parsing file is simple, but it is performance killer: encode to json, IO write, read, decode from json, create dict objects, code complexity++ **VS** pass reference pointer to caller. (seems weird to do this within the same python process)

If there will be a new API that will improve a way how to work with operator-courier as python module from services, I'm fine implementing a *temporary* workaround with JSON file read.

chmeliik:
@kevinrizza The immediate use case is that it gives you a way to access the validation info other than *convert the dict into a json, write the json to a file, read the json from the file, convert it back to a dict*. While that does work, it feels pretty bad to have to do that every time.

> The `operator-courier` right now is intended to simply be a tool that can verify and push these bundles at a high level.

As is the `OperatorBundle` class. It only provides an object-oriented interface to the functionality that `operator-courier` already provides. The only only "feature" it has over the current API is, as stated above, access to the dict with validation results. And I would argue that providing the functionality through this model is neater than having a `build_verify_and_push()` function that does everything. This way you can build the bundle, then check if it is valid, show the validation output to a user, write the bundle or the validation output to a file, etc.

> Even if we were to go to such a model, I would prefer that the `bundle` class itself to be viewed as `__private` implementation details.

I'm not sure what you mean by this. Do you not want the `OperatorBundle` class to be exported by the `api` module? That is understandable. Other than that, I don't see a reason to hide any of the class methods, unless there is a chance you would stop providing validation results in the future.
Conversely, if you do decide to add functionality in the future, it would be as simple as adding a method to the `OperatorBundle` class, rather than having a `build_verify_push_and_do_this_other_thing()` method.

As for the bundle format changes, that is unfortunate and understandable. However, I don't see how having a bundle class would make the transition that much harder. Or, as Martin already said, you could have this be the class for old bundles and make a new one for the new format.

Anyway, that's just my 2 cents.

kevinrizza:
@MartinBasti @chmeliik 

Please imagine the `operator-courier verify` to be tangentially similar to the `docker build` cli, and the `operator-courier push` to be similar to the `docker push` cli:

When you build a docker image, `docker build` does not return the image itself. It builds a blob and pushes it somewhere on disk, and then gives you the knowledge of how to access it. We could do something similar here if we wanted to inspect the output of the build.

When you push an image, docker definitely doesn't return the image. I do not believe we ever want courier to be in the business of returning anything more than trivial metadata as part of a push.

IMO the API should provide the same abstraction as the CLI -- the goal of this tool should be to take input in the form of files (whether that is yaml, source code, Dockerfiles, binaries, whatever) and use them to generate and verify some bundle, then push that bundle if so desired. We have a 1:1 relationship between the CLI and the API for a reason, because we only want to support those functions long term, especially as and when the format changes. Creating a bundle type creates exactly the sort of abstraction we want to avoid. My personal thought in the future is that these "bundles" will actually just become `operator-registry` images that we push to quay or any other container repository. If we do that, the concept of a "bundle" goes away anyway and that entire api doesn't make sense anymore. If we minimize what it actually does in terms of features and options to just the stated goals, it would allow us to seamlessly move to such a format in incremental changes without fundamentally redefining the APIs every time:

 "Take the things and make a new thing" -> "Tell me if something is wrong or invalid" -> "Push the thing"

That is a much simpler worldview. And that is much more achievable if we do not include complex implementation details into the use of the API. Having a bundle class that publicly returns some object that itself has methods to mutate the state of it is definitely outside the scope of those stated goals. I wouldn't expect anyone running the CLI tool to understand those implementation details, and by extension I feel the same way about the API.

chmeliik:
@kevinrizza Wait, so if I understand this correctly, the return value from `api.build_and_verify()`, which currently returns the formatted bundle, is not meant to be used from the outside? If so, would it be possible to simply have `build_and_verify()` return the validation info instead, and to also return the same info from `build_verify_and_push()`? Or would that also be outside the intended scope?

kevinrizza:
@chmeliik No, I'm not saying that what is returned from build_and_verify() can't be used from the outside. I'm saying that all it returns is a map which is equivalent to a yaml file. And, that the yaml it returns should not be considered an API -- which is to say, that none of the fields returned by such a blob should be expected to always be there. Similarly, it should not be expected that the blob it returns now will always be in the same shape or even of the same type. We only guarantee that such a blob exists.

That is why I purposely chose not to make this blob an object oriented class. The state will mutate significantly enough in the future that expecting this type to be reasoned about (with things like most methods and fields) is difficult right now. In the future, if the format becomes much more stable, we could move to such a design.

I think in the short term, since the only guarantee is that "what is returned by `build_and_verify()` is a blob which represents a bundle and could be written to disk in some form, that we can still return it as we do now. We could even add a flag to the CLI to write it to disk in a preferred folder path. But we shouldn't make it any sort of well defined and public type that is expected to be reasoned about. If that is done in the short term, it is something that will definitely break with a major version update to the courier.

chmeliik:
@kevinrizza Ok, that is fair. I won't waste more of your time then, feel free to close this PR.

kevinrizza:
@chmeliik It's no problem. Thank you for the PR and the back and forth. I think part of the issue lies in the fact that this project should do a better job of defining its stated purpose and goals going forward. There is some very valuable feedback here, and hopefully in the coming weeks we can attempt to address a lot of this.

PR #80, opened 2019-03-21
Use argparse subparsers for subcommands

### LINKED ISSUE #77: document --ui_validate_io switch
--ui_validate_io was asked to be used to verify data for operatorhub.io, but it isn't documented in readme or in the --help output.

csomh:
Makes argument parsing more resilient, by handling
subcommands with argparse provided and recommended [subparsers][1],
instead of the current custom manipulation of argv.

Removes the need to manually update and format the top-level
help message: this is going to be composed from the help strings
of the subcommands by argparse.

Fixes the issue of subcommand's usage string not displaying the
subcommand.

Removes the need to explicitly check for incorrect subcommands.

Enables extending the CLI with top-level optional arguments, without the
need of adjusting argv parsing.

[1]: https://docs.python.org/3/library/argparse.html#sub-commands

### DIFF
--- a/operatorcourier/cli.py
+++ b/operatorcourier/cli.py
@@ -28,72 +28,114 @@ class _CliParser():
         """
         parser = argparse.ArgumentParser(
             description='Build, verify and push operator bundles into '
-                        'external app registry',
-            usage='''operator-courier <command> [<args>]
+                        'external app registry')
 
-These are the commands you can use:
-    verify      Create a bundle and test it for correctness.
-    push        Create a bundle, test it, and push it to an app registry.
-    nest        Take a flat to-be-bundled directory and version nest it.
-    flatten     Create a flat directory from versioned operator bundle yaml files.
-''')
         try:
             __version__ = pkg_resources.get_distribution('operator-courier').version
         except Exception:
             __version__ = 'unknown'
 
-        parser.add_argument('command', help='Subcommand to run')
-        parser.add_argument('-v', '--version',
-                            help='Show the current version of operator-courier',
-                            action='version', version=__version__)
+        parser.add_argument(
+            '-v', '--version',
+            help='Show the current version of operator-courier',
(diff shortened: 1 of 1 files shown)


kevinrizza:
What is the advantage of using `subcommand` versus the current implementation?

Also, https://github.com/operator-framework/operator-courier/issues/77 should be solved by a separate pr. Please do not co-mingle multiple changes in a single pull request.

csomh:
[Subparsers][1] are the argparse provided way to handle subcommands. Switching to this was easier than figuring out the bug in the custom implementation, which was causing #77 or led to an incorrect usage text to be displayed for subcommands.

Sorry, the commit message didn't clarify this. Updated it, together with the initial PR comment.

[1]: https://docs.python.org/3/library/argparse.html#sub-commands

kevinrizza:
How do you know that there was a bug at all? It seems like the command has description output, and I'm not sure that your change fixed any bug at all. It seems more likely to me that the issue in question was describing a perceived lack of documentation about what the flag outputs (aside from the very basic description of the flag).

That could be a separate discussion, but I don't think this PR fixes the issue.

And, again, I still don't see what using `subcommand`s contributes here. Is there some design limitation to the current implementation (aside from not using some specific sub library)? What does moving to using `subcommand` actually give us that we don't have already? If there is a technical argument to be made for them, I am certainly interested in hearing it.

csomh:
Oh, ok, I see now. You're right.

MartinBasti:
This PR is still valid (but must be updated), There were already several issues caused by not using subparsers.

* https://github.com/operator-framework/operator-courier/pull/85
* https://github.com/operator-framework/operator-courier/pull/89
* https://github.com/operator-framework/operator-courier/issues/88

csomh:
Updated.

MartinBasti:
works for me

PR #82, opened 2019-03-21
Update tox for integration tests and add tests for push

jeremylinlin:
- Update tox config to run integration tests
- Add a script that runs the set of integration tests based on Travis build triggering event
- Add integration tests for `push` command

### DIFF
--- a/run_integration_tests.sh
+++ b/run_integration_tests.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+echo "Running integration tests triggered by $TRAVIS_EVENT_TYPE..."
+
+pytest tests/integration/test_verify.py
+
+# only runs if the Travis build is triggered by a cronjob, or
+# $QUAY_NAMESPACE and $QUAY_ACCESS_TOKEN is set locally
+if [[ $TRAVIS_EVENT_TYPE == 'cron' ]] || (! [[ -z $QUAY_NAMESPACE ]] && ! [[ -z $QUAY_ACCESS_TOKEN ]]); then
+  pytest tests/integration/test_push.py
+fi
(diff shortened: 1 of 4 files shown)


run_integration_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+echo "Running integration tests triggered by $TRAVIS_EVENT_TYPE..."
+if [[ $TRAVIS_EVENT_TYPE == 'cron' ]]; then
+  echo 'the integration test command for push will be enabled after env vars are set up for Travis CI'
+#  pytest tests/integration/test_push.py
```
kevinrizza: We probably shouldn't merge something that references a code path that doesn't exist yet.

jeremylinlin: @kevinrizza the `echo` command serves as a placeholder and the actual command is commented out.

I wanted to make the PRs more separated since they serve as two purposes. And it currently won't break anything and we don't have cron jobs yet.

Do you still want me to put them all in one PR?

kevinrizza: @jeremylinlin I just don't see why we want to merge this first, given that it doesn't do anything. Pull requests should be singular in purpose, but if your purpose is to add new tests, and those tests require stuff around them to work correctly, they should be in one pr. If you prefer, you could make it two commits in the same pr.

kevinrizza:
@jeremylinlin is the infrastructure set up for these yet?

jeremylinlin:
> @jeremylinlin is the infrastructure set up for these yet?

@kevinrizza  Not yet. I'll talk to Aravindh when he comes back. I have tested it on my own CI environment though.

kevinrizza:
@jeremylinlin In that case, let's hold off on reviewing/merging this pr until we have it set up. We should definitively run a real end to end test on our real CI as a proof of concept to make sure that no additional changes are needed.

Issue #74, opened 2019-03-18
When user provides invalid files, the error does not mention which files are invalid and why


scoheb:
(tmp-XVr6zr33) shebert@localhost/tmp/scott:$ operator-courier -v
1.2.0
(tmp-XVr6zr33) shebert@localhost/tmp/scott:$ operator-courier verify  .
ERROR:operatorcourier.identify:Courier requires valid CSV, CRD, and Package files
Courier requires valid CSV, CRD, and Package files

awgreene:
Hello @scoheb. This error would result from the following conditions:
1. A file in the directory is not a valid yaml file.
2. If the directory only contains valid yaml files, then this log would be produced if none of the following conditions are met:
- The `packageName` field  doesn't exist.
- The `kind` is not set to `ClusterServiceVersion` or `CustomResourceDefintion`.

The code that produced this log can be seen [here](https://github.com/operator-framework/operator-courier/blob/master/operatorcourier/identify.py).

We could update the message to report that one of the conditions above was not met.

scoheb:
@awgreene The above text would be good to output to the user. Even better would be your suggestion of `We could update the message to report that one of the conditions above was not met.`

brianwcook:
request we say -which- file is not a valid yaml file if possible

brianwcook:
also i assume you are using a custom config for linting?  I get hard errors from yammlint like

23:81     error    line too long (3902 > 80 characters)  (line-length)

but obviously this is required due to the base64 encoded images in the yaml.  So I think we need detailed yaml lint output reported so we know what issues actually are considered errrors by operator-courier.

PR #83, opened 2019-03-22
Start using JSON Schema for validation

csomh:
Refactor CRD and CSV validation, and use JSON Schema
instead of nested if-s.

This should make future validation easier.

TODO:
- [x] write some documentation explaining all this.


csomh:
This should resolve #17 and #39.

kevinrizza:
@csomh Just wanted to say good work on this! It seems like it satisfies a lot of the basic cases, and allows us to just codify the more complex comparisons. I would like to take some time and ensure that all of the use cases that were covered previously are still working here -- but in general I do like this approach.

One thing I want to point out is that this does change the formatting of most/all of the log messages, which to me is a breaking change -- so we would have to either update this PR to retain the old format of the logs or we would need to bump the major version. I would prefer the former.

scoheb:
FYI @kevinrizza as one of the consumers of courier...CVP does not rely on the format of the log messages.

csomh:
@kevinrizza thanks!

As for the format of log messages: would it be possible to sync releasing this change with the future fix for #74, which will probably also change the format of the log messages?

I'm asking this, because keeping the old format of the messages requires parsing the `error.message` from `jsonschema` and reconstructing the current error messages. But then, the current error messages resulting from validation issues will probably not match in style the new error messages that will pop up due to fine tuning the schemas.

I would rather stick to the current _concatenate `error.absolute_path` with `error.message`_ approach, as it's short, simple and provides a consistent style on the long run.

kevinrizza:
@csomh Agreed, if we are going to make format changes to all of the log messages, my preference would be to make a more sweeping change that would give more detail as to what csv or crd is causing the error. I'm not opposed to the changes to the logger in principle, just to the timing of them. If that is directly tied to this PR because of technical reasons, then lets hold this pr until we make that change as well.

csomh:
I think this isn't relevant anymore :slightly_smiling_face: Closing.

PR #85, opened 2019-03-25
Add flatten command

jeremylinlin:
A new operator-courier CLI command called `flatten` which, given a directory of operator bundles, it will extract the versioned csvs and latest version of each crd along with the package file and create a new flat directory of yaml files.

### DIFF
--- a/operatorcourier/api.py
+++ b/operatorcourier/api.py
@@ -15,6 +15,7 @@ from operatorcourier.validate import ValidateCmd
 from operatorcourier.push import PushCmd
 from operatorcourier.format import format_bundle
 from operatorcourier.nest import nest_bundles
+from operatorcourier.flatten import flatten_bundles
 from operatorcourier.errors import OpCourierBadBundle
 
 logger = logging.getLogger(__name__)
@@ -136,3 +137,18 @@ def nest(source_dir, registry_dir):
 
     with TemporaryDirectory() as temp_dir:
         nest_bundles(yaml_files, registry_dir, temp_dir)
+
+
+def flatten(source_dir, dest_dir):
+    """
+    Given a directory containing different versions of operator bundles
+    (CRD, CSV, package) in separate version directories, this function
+    copies all files needed to the flattened directory. It is the inverse
+    of the `nest` function.
+
+    :param source_dir: the directory containing different versions
+    of operator bundles (CRD, CSV, package) in separate version directories
+    :param dest_dir: the flattened directory path where all necessary files are copied
+    """
+    os.makedirs(dest_dir, exist_ok=True)
+    flatten_bundles(source_dir, dest_dir)
--- a/operatorcourier/cli.py
+++ b/operatorcourier/cli.py
@@ -35,6 +35,7 @@ These are the commands you can use:
     verify      Create a bundle and test it for correctness.
     push        Create a bundle, test it, and push it to an app registry.
     nest        Take a flat to-be-bundled directory and version nest it.
+    flatten     Create a flat directory from versioned operator bundle yaml files.
 ''')
         try:
             __version__ = pkg_resources.get_distribution('operator-courier').version
@@ -113,3 +114,24 @@ These are the commands you can use:
 
         args, leftovers = parser.parse_known_args(sys.argv[2:])
         api.nest(args.source_dir, args.registry_dir)
+
+    # Parse the flatten command
+    def flatten(self):
+        parser = argparse.ArgumentParser(
+            usage='operator-courier flatten [-h] source_dir dest_dir',
+            description='Given a directory with different versions of '
+                        'operator bundles (CRD, CSV, package), this command extracts '
+                        'versioned CSVs and the latest version of each CRD along with '
+                        'the package file and creates a new flat directory '
+                        'of yaml files. See https://github.com/operator-framework/'
+                        'operator-registry#manifest-format to find out more about '
+                        'how nested bundles should be structured.')
+        parser.add_argument('source_dir',
+                            help='Path of the source directory that contains different '
+                                 'versions of operator bundles (CRD, CSV, package)')
+        parser.add_argument('dest_dir',
+                            help='The new flat directory that contains '
+                                 'extracted bundle files')
+
+        args, leftovers = parser.parse_known_args(sys.argv[2:])
+        api.flatten(args.source_dir, args.dest_dir)
--- a/operatorcourier/flatten.py
+++ b/operatorcourier/flatten.py
@@ -1,0 +1,170 @@
+import logging
+import os
+from yaml import safe_load, MarkedYAMLError
+from typing import Dict, Tuple
+from shutil import copyfile
+import semver
+from operatorcourier import identify
+from operatorcourier import errors
+
+logger = logging.getLogger(__name__)
+
+
+def flatten_bundles(source_dir: str, dest_dir: str):
+    file_paths_to_copy = get_flattened_files_info(source_dir)
+    for (src_file_path, new_file_name) in file_paths_to_copy:
+        copyfile(src_file_path, os.path.join(dest_dir, new_file_name))
+
+
+def get_flattened_files_info(source_dir: str) -> [(str, str)]:
+    """
+    :param source_dir: Path of the directory containing different versions
+    of operator bundles (CRD, CSV, package) in separate version directories
+    :return: A list of tuples where in each tuple, the first element is
+    the source file path to be copied to the flattened directory, and the second is
+    the new file name to be used in the file copy. The function returns an empty list if
+    the source_dir is already flat
+    """
+
+    # extract package file and version folders from source_dir
+    _, folder_names, file_names = next(os.walk(source_dir))
+    if not folder_names:
+        logger.info('The source directory is already flat.')
+        return []
+
+    file_paths_to_copy = []  # [ (SRC_FILE_PATH, NEW_FILE_NAME) ]
+
+    crd_dict = {}  # { CRD_NAME => (VERSION, CRD_PATH) }
+    csv_paths = []
+
+    for version_folder_name in folder_names:
+        parse_version_folder(source_dir, version_folder_name, csv_paths, crd_dict)
+
+    # add package in source_dir
+    package_path = get_package_path(source_dir, file_names)
+    file_paths_to_copy.append((package_path, os.path.basename(package_path)))
+
+    # add all CRDs with the latest version
+    for _, crd_path in crd_dict.values():
+        file_paths_to_copy.append((crd_path, os.path.basename(crd_path)))
+
+    # add all CSVs
+    for csv_file_name, csv_entries in create_csv_dict(csv_paths).items():
+        for (version, csv_path) in csv_entries:
+            basename, ext = os.path.splitext(csv_file_name)
+            file_paths_to_copy.append((csv_path, f'{basename}-v{version}{ext}'))
+
+    return file_paths_to_copy
+
+
+def parse_version_folder(base_dir: str, version_folder_name: str,
+                         csv_paths: list, crd_dict: Dict[str, Tuple[str, str]]):
+    """
+    Parse the version folder of the bundle and collect information of CSV and CRDs
+    in the bundle
+
+    :param base_dir: Path of the base directory where the version folder is located
+    :param version_folder_name: The name of the version folder containing bundle files
+    :param csv_paths: A list of CSV file paths inside version folders
+    :param crd_dict: dict that contains CRD info collected from different version folders,
+    where the key is the CRD name, and the value is a tuple where the first element is
+    the version of the bundle, and the second is the path of the CRD file
+    """
+    # parse each version folder and parse CRD, CSV files
+    try:
+        semver.parse(version_folder_name)
+    except ValueError:
+        logger.warning("Ignoring %s as it is not a valid semver. "
+                       "See https://semver.org for the semver specification.",
+                       version_folder_name)
+        return
+
+    logger.info('Parsing folder: %s...', version_folder_name)
+
+    contains_csv = False
+    version_folder_path = os.path.join(base_dir, version_folder_name)
+
+    for item in os.listdir(os.path.join(base_dir, version_folder_name)):
+        item_path = os.path.join(version_folder_path, item)
+
+        if not os.path.isfile(item_path):
+            logger.warning('Ignoring %s as it is not a regular file.', item)
+            continue
+        if not is_yaml_file(item_path):
+            logging.warning('Ignoring %s as the file does not end with .yaml or .yml',
+                            item_path)
+            continue
+
+        with open(item_path, 'r') as f:
+            file_content = f.read()
+
+        yaml_type = identify.get_operator_artifact_type(file_content)
+
+        if yaml_type == 'ClusterServiceVersion':
+            contains_csv = True
+            csv_paths.append(item_path)
+        elif yaml_type == 'CustomResourceDefinition':
+            try:
+                crd_name = safe_load(file_content)['metadata']['name']
+            except MarkedYAMLError:
+                msg = "Courier requires valid input YAML files"
+                logger.error(msg)
+                raise errors.OpCourierBadYaml(msg)
+            except KeyError:
+                msg = f'{item} is not a valid CRD file as "metadata.name" ' \
+                      f'field is required'
+                logger.error(msg)
+                raise errors.OpCourierBadBundle(msg)
+            # create new CRD type entry if not found in dict
+            if crd_name not in crd_dict:
+                crd_dict[crd_name] = (version_folder_name, item_path)
+            # update the CRD type entry with the file with the newest version
+            elif semver.compare(version_folder_name, crd_dict[crd_name][0]) > 0:
+                crd_dict[crd_name] = (crd_dict[crd_name][0], item_path)
+
+    if not contains_csv:
+        msg = 'This version directory does not contain any valid CSV file.'
+        logger.error(msg)
+        raise errors.OpCourierBadBundle(msg)
+
+
+def get_package_path(base_dir: str, file_names_in_base_dir: list) -> str:
+    packages = []
+
+    # add package file to file_paths_to_copy
+    # only 1 package yaml file is expected in file_names
+    for file_name in file_names_in_base_dir:
+        file_path = os.path.join(base_dir, file_name)
+        if not is_yaml_file(file_path):
+            logging.warning('Ignoring %s as the file does not end with .yaml or .yml',
+                            file_path)
+            continue
+
+        with open(file_path, 'r') as f:
+            file_content = f.read()
+        if identify.get_operator_artifact_type(file_content) != 'Package':
+            logger.warning('Ignoring %s as it is not a valid package file.', file_name)
+        elif not packages:
+            packages.append(file_path)
+        else:
+            msg = f'The input source directory expects only 1 valid package file.'
+            logger.error(msg)
+            raise errors.OpCourierBadBundle(msg)
+
+    return packages[0]
+
+
+# parse all CSVs and ensure those with same names are handled
+def create_csv_dict(csv_paths: list) -> dict:
+    csv_dict = {}  # { CSV_FILE_NAME => [ (v1, csv_path_1), ... , (vn, csv_path_n) ] }
+    for csv_path in csv_paths:
+        version_folder_path, csv_file_name = os.path.split(csv_path)
+        version = os.path.basename(version_folder_path)
+        val = (version, csv_path)
+        csv_dict.setdefault(csv_file_name, []).append(val)
+    return csv_dict
+
+
+def is_yaml_file(file_path: str) -> bool:
+    yaml_ext = ['.yaml', '.yml']
+    return os.path.splitext(file_path)[1] in yaml_ext
--- a/Pipfile
+++ b/Pipfile
@@ -10,6 +10,6 @@ requests = "*"
 pytest = "*"
 pyyaml = "*"
 twine = "*"
-
+semver = "*"
 
 [dev-packages]
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "18523475689e3183d2da8dfea9f6f6e55520811e8a2812d7415feb4738a08cc9"
+            "sha256": "5d7ac951678f5b9207ea989f0dc2e89ebecf9348d1558cebfa384e8635554890"
         },
         "pipfile-spec": 6,
         "requires": {},
@@ -23,10 +23,10 @@
         },
         "attrs": {
             "hashes": [
-                "sha256:10cbf6e27dbce8c30807caf056c8eb50917e0eaafe86347671b57254006c3e69",
-                "sha256:ca4be454458f9dec299268d472aaa5a11f67a4ff70093396e1ceae9c76cf4bbb"
+                "sha256:69c0dbf2ed392de1cb5ec704444b08a5ef81680a61cb899dc08127123af36a79",
+                "sha256:f0b870f674851ecbfbbbd364d6b5cbdff9dcedbc7f3f5e18a6891057f21fe399"
             ],
-            "version": "==18.2.0"
+            "version": "==19.1.0"
         },
         "bleach": {
             "hashes": [
@@ -37,10 +37,10 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:47f9c83ef4c0c621eaef743f133f09fa8a74a9b75f037e8624f83bd1b6626cb7",
-                "sha256:993f830721089fef441cdfeb4b2c8c9df86f0c63239f06bd025a76a7daddb033"
+                "sha256:59b7658e26ca9c7339e00f8f4636cdfe59d34fa37b9b04f6f9e9926b3cece1a5",
+                "sha256:b26104d6835d1f5e49452a26eb2ff87fe7090b89dfcaee5ea2212697e1e1d7ae"
             ],
-            "version": "==2018.11.29"
+            "version": "==2019.3.9"
         },
         "chardet": {
             "hashes": [
@@ -48,6 +48,13 @@
                 "sha256:fc323ffcaeaed0e0a02bf4d117757b98aed530d9ed4531e3e15460124c106691"
             ],
             "version": "==3.0.4"
+        },
+        "decorator": {
+            "hashes": [
+                "sha256:86156361c50488b84a3f148056ea716ca587df2f0de1d34750d35c21312725de",
+                "sha256:f069f3a01830ca754ba5258fde2278454a0b5b79e0d7f5c13b3b97e57d4acff6"
+            ],
+            "version": "==4.4.0"
         },
         "docutils": {
             "hashes": [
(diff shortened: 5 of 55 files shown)


operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
...
+def get_flattened_files(source_dir: str) -> list:
+    """
+    :param source_dir: the directory containing different versions
+    of operator bundles (CRD, CSV, package) in separate version directories
+    :return: a list of files needed to be copied to the flattened directory
+    """
+
+    # extract package file and version folders from source_dir
+    file_names, folder_names = [], []
+    for item in os.listdir(source_dir):
```
MartinBasti: wouldn't be easier just use `os.walk` https://www.tutorialspoint.com/python/os_walk.htm?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
...
+        item_path = f'{source_dir}/{item}'
```
MartinBasti: This is not a safe way how to construct file paths, use `os.path.join` or `Path` object

operatorcourier/api.py
```diff
@@ -136,3 +138,21 @@ def nest(source_dir, registry_dir):
...
+    Given a directory containing different versions of operator bundles
+    (CRD, CSV, package) in separate version directories, this function
+    copies all files needed to the flattened directory
+
+    :param source_dir: the directory containing different versions
+    of operator bundles (CRD, CSV, package) in separate version directories
+    :param dest_dir: the flattened directory path where all necessary files are copied
+    """
+    if not os.path.exists(dest_dir):
+        os.makedirs(dest_dir)
```
csomh: `if` could be removed:

```python
os.makedirs(dest_dir, exists_ok=True)
```

operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
...
+    # copy all CRDs with the latest version
+    for crd_name, (_, crd_path) in crd_dict.items():
+        file_paths_to_copy.append(crd_path)
+
+    return file_paths_to_copy
+
+
+def newer_semantic_version(input_semver_str: str, current_latest_semver_str: str) -> bool:
+    """
+    Given two VALID semantic version strings
```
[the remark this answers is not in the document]
MartinBasti: +:100: in PR #68 semver module was already approved

operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
...
+                logger.warning('Ignoring %s as it is not a regular file.', item)
+
+            with open(item_path, 'r') as f:
+                file_content = f.read()
+
+            yaml_type = identify.get_operator_artifact_type(file_content)
+
+            if yaml_type == 'ClusterServiceVersion':
+                contains_csv = True
+                file_paths_to_copy.append(item_path)
```
MartinBasti: What happen when CSV files have the same name? Only latest CRD is used, but CSV is copied in pseudo-random order and overwriting previous CSVs if they have the same name. Shouldn't this ensure that latest CSV is copied as last? (no and expert in CSV/CRD please correct me)

operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
...
+            elif yaml_type == 'CustomResourceDefinition':
+                crd_name = safe_load(file_content)['metadata']['name']
```
MartinBasti: Don't you want to raise here `OpCourier*` errors instead `ParseError/KeyError`

operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
...
+        with open(file_path, 'r') as f:
+            file_content = f.read()
+        if identify.get_operator_artifact_type(file_content) != 'Package':
+            logger.warning('Ignoring %s as it is not a valid package file.', file_name)
+        else:
+            package_paths.append(file_path)
+
+    n_packages = len(package_paths)
+    if n_packages != 1:
+        logger.error('The input source directory expects 1 valid package file, '
```
MartinBasti: `raise` something here ?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
...
+                crd_name = safe_load(file_content)['metadata']['name']
+                # create new CRD type entry if not found in dict
+                if crd_name not in crd_dict:
+                    crd_dict[crd_name] = (version_folder_name, item_path)
+                # update the CRD type entry with the file with the newest version
+                elif newer_semantic_version(version_folder_name, crd_dict[crd_name][0]):
+                    crd_dict[crd_name] = (crd_dict[crd_name][0], item_path)
+
+        if not contains_csv:
+            logger.error('This version directory does not contain any valid CSV file.')
```
MartinBasti: `raise` something here?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,131 @@
+import logging
+import os
+import re
+from yaml import safe_load
+import operatorcourier.identify as identify
+
+logger = logging.getLogger(__name__)
+
+
+def get_flattened_files(source_dir: str) -> list:
```
MartinBasti: this function is too long, should be split to parts to increase readability

operatorcourier/flatten.py
kevinrizza: All of the CSVs are required. So the only options are to either: enforce that CSVs have different file names (which is the current standard https://github.com/operator-framework/operator-registry#manifest-format) or if duplicates are found rename the files when they are copied by appending the version to the file name or something similar.

operatorcourier/flatten.py
kevinrizza: +1

operatorcourier/flatten.py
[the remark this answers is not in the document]
kevinrizza: Yes please, that significantly simplifies the code:

https://python-semver.readthedocs.io/en/latest/api.html#semver.compare

kevinrizza:
Could we include a link to the nested bundle docs somewhere, at least in the help message?

https://github.com/operator-framework/operator-registry#manifest-format

operatorcourier/flatten.py
MartinBasti: Because we have version in dirname, I prefer to add it to CSV filename

MartinBasti:
Also would be nice if dir structure is already flat to not raise an error but log only INFO message

csomh:
> Also would be nice if dir structure is already flat to not raise an error but log only INFO message

To clarify this a little bit: we discussed with @ralphbean that in order to avoid confusion among OMPS users, the service should be able to handle both flattened and nested content, using the same endpoint. To be able to do this, OMPS either

- should be able to tell the difference between the two, and probably should rely on Operator Courier to get this information. This would require an `is_flat` or something similar API.

OR

- `flatten` should do nothing if called on a "flat" tree.

jeremylinlin:
@MartinBasti @csomh @kevinrizza Thank you guys very much for the very useful comments. I have addressed all of them and please review my changes when you are free.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+                crd_name = safe_load(file_content)['metadata']['name']
+            except MarkedYAMLError:
+                msg = "Courier requires valid input YAML files"
+                logger.error(msg)
+                raise errors.OpCourierBadYaml(msg)
+            # create new CRD type entry if not found in dict
+            if crd_name not in crd_dict:
+                crd_dict[crd_name] = (version_folder_name, item_path)
+            # update the CRD type entry with the file with the newest version
+            elif semver.compare(version_folder_name, crd_dict[crd_name][0]) == 1:
```
csomh: How about using `> 0` for comparison, to match the docs for `semver.compare`:

> Returns: The return value is negative if ver1 < ver2, zero if ver1 == ver2 and strictly positive if ver1 > ver2

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+                crd_dict[crd_name] = (crd_dict[crd_name][0], item_path)
+
+    if not contains_csv:
+        msg = 'This version directory does not contain any valid CSV file.'
+        logger.error(msg)
+        raise errors.OpCourierBadBundle(msg)
+
+
+def get_package_file(base_dir: str, file_names_in_base_dir) -> (str, str):
+    packages = []  # [ (PACKAGE_FILE_PATH, PACKAGE_FILE_NAME) ]
```
csomh: Why store both? The filename can be easily obtained with `os.path.basename()`.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+        file_path = os.path.join(base_dir, file_name)
+        with open(file_path, 'r') as f:
+            file_content = f.read()
+        if identify.get_operator_artifact_type(file_content) != 'Package':
+            logger.warning('Ignoring %s as it is not a valid package file.', file_name)
+        else:
+            packages.append((file_path, file_name))
+
+    n_packages = len(packages)
+    if n_packages != 1:
```
csomh: Wouldn't be easier to abort as soon as the second package file is found, instead of reading all, just to throw them away?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+        raise errors.OpCourierBadBundle(msg)
+    else:
+        return packages[0][0], packages[0][1]
+
+
+# parse all CSVs and ensure those with same names are handled
+def create_csv_dict(csv_paths: list) -> dict:
+    csv_dict = {}  # { CSV_FILE_NAME => [ (v1, csv_path_1), ... , (vn, csv_path_n) ] }
+    for csv_path in csv_paths:
+        version, csv_file_name = csv_path.split('/')[-2:]
```
csomh:
```suggestion
        version, csv_file_name = os.path.split(csv_path)
```

Please reserve string operations on paths for file-systems experts :D

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+        val = (version, csv_path)
+        if csv_file_name in csv_dict:  # duplicate CSVs found in different versions
```
csomh: For this `if.. else` block there is a Pythonic way:

```suggestion
        csv_dict.setdefault(csv_file_name, []).append(val)
```

Read [setdefault](https://docs.python.org/3/library/stdtypes.html#dict.setdefault).

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+    of operator bundles (CRD, CSV, package) in separate version directories
+    :return: A list of tuples where in each tuple, the first element is
+    the source file path to be copied to the flattened directory, and the second is
+    the new file name to be used in the file copy. The function returns an empty list if
+    the source_dir is already flat
+    """
+
+    # extract package file and version folders from source_dir
+    _, folder_names, file_names = next(os.walk(source_dir))
+    if len(folder_names) == 0:
```
csomh:
```suggestion
    if not folder_names:
```

is just as good.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
+import logging
+import os
+from yaml import safe_load, MarkedYAMLError
+import semver
+import ntpath
```
csomh: Please, no! This is a platform specific module for Windows. `os.path` will do the job.

> Since different operating systems have different path name conventions, there are several versions of this module in the standard library. The os.path module is always the path module suitable for the operating system Python is running on, and therefore usable for local paths.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+
+    for version_folder_name in folder_names:
+        parse_version_folder(source_dir, version_folder_name, csv_paths, crd_dict)
+
+    # 1. add package in source_dir
+    package_path, package_name = get_package_file(source_dir, file_names)
+    file_paths_to_copy.append((package_path, package_name))
+
+    # 2. add all CRDs with the latest version
+    for crd_name, (_, crd_path) in crd_dict.items():
```
csomh:
```suggestion
    for _, crd_path in crd_dict.values():
```

If the key is not used, ask only for the values.

operatorcourier/cli.py
```diff
@@ -113,3 +114,24 @@ def nest(self):
 
         args, leftovers = parser.parse_known_args(sys.argv[2:])
         api.nest(args.source_dir, args.registry_dir)
+
+    # Parse the flatten command
+    def flatten(self):
```
csomh: I think something is missing from the usage string of `flatten -h`:

```bash
$ operator-courier flatten -h
usage: operator-courier [-h] source_dir dest_dir
```

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+        file_paths_to_copy.append((crd_path, ntpath.basename(crd_path)))
+
+    # 3. add all CSVs
+    for csv_file_name, csv_entries in create_csv_dict(csv_paths).items():
+        if len(csv_entries) == 1:
+            file_paths_to_copy.append((csv_entries[0][1], csv_file_name))
+            continue
+        # duplicate CSVs found in different versions
+        for (version, csv_path) in csv_entries:
```
csomh: Why not do this transformation also when `len(csv_entries) == 1`? And then the if-block above could be deleted.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+            basename, ext = os.path.splitext(csv_file_name)
+            file_paths_to_copy.append((csv_path, f'{basename}_v{version}{ext}'))
```
csomh:
```suggestion
            file_paths_to_copy.append((csv_path, f'{basename}.v{version}{ext}'))
```

Would it be more consistent to use `.` as a separator here?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+    :return: A list of tuples where in each tuple, the first element is
+    the source file path to be copied to the flattened directory, and the second is
+    the new file name to be used in the file copy. The function returns an empty list if
+    the source_dir is already flat
+    """
+
+    # extract package file and version folders from source_dir
+    _, folder_names, file_names = next(os.walk(source_dir))
+    if len(folder_names) == 0:
+        logger.info('The source directory is already flat. Exiting the program...')
```
MartinBasti: Don't exit anything here, this is part of API, I'm sure I don't want SysOps to see this message in service logs. Drop `Exiting the program...` part please.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+
+        yaml_type = identify.get_operator_artifact_type(file_content)
+
+        if yaml_type == 'ClusterServiceVersion':
+            contains_csv = True
+            csv_paths.append(item_path)
+        elif yaml_type == 'CustomResourceDefinition':
+            try:
+                crd_name = safe_load(file_content)['metadata']['name']
+            except MarkedYAMLError:
```
MartinBasti: How about adding `KeyError` here too?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+        # duplicate CSVs found in different versions
+        for (version, csv_path) in csv_entries:
+            basename, ext = os.path.splitext(csv_file_name)
+            file_paths_to_copy.append((csv_path, f'{basename}_v{version}{ext}'))
+
+    return file_paths_to_copy
+
+
+def parse_version_folder(base_dir: str, version_folder_name: str,
+                         csv_paths: list, crd_dict: (str, str)):
```
MartinBasti: is `(str, str)` the right way how to describe dict in typing? Shouldn't be this `Dict[str, Tuple[str, str]]` ?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+    packages = []  # [ (PACKAGE_FILE_PATH, PACKAGE_FILE_NAME) ]
+
+    # add package file to file_paths_to_copy
+    # only 1 package yaml file is expected in file_names
+    for file_name in file_names_in_base_dir:
+        file_path = os.path.join(base_dir, file_name)
+        with open(file_path, 'r') as f:
+            file_content = f.read()
+        if identify.get_operator_artifact_type(file_content) != 'Package':
+            logger.warning('Ignoring %s as it is not a valid package file.', file_name)
```
MartinBasti: Directory should be used too, is hard to identify which file when files are nested in dirs.

operatorcourier/flatten.py
MartinBasti: also I don't see `mypy` enabled in `tox`

operatorcourier/flatten.py
kevinrizza: I would actually prefer ` - ` to ` . ` or ` _ `, since the underscore character can be problematic and the period character is used in the version string itself.

tox.ini
```diff
@@ -2,7 +2,9 @@
 envlist = py36,py37,flake8
 
 [testenv]
-deps = .[test]
+deps =
+    .[test]
+    semver
```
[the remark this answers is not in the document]
kevinrizza: +1

operatorcourier/api.py
```diff
@@ -136,3 +138,21 @@ def nest(source_dir, registry_dir):
...
+    :param source_dir: the directory containing different versions
+    of operator bundles (CRD, CSV, package) in separate version directories
+    :param dest_dir: the flattened directory path where all necessary files are copied
+    """
+    os.makedirs(dest_dir, exist_ok=True)
+
+    file_paths_to_copy = get_flattened_files_info(source_dir)
+    if file_paths_to_copy:
+        for (src_file_path, new_file_name) in file_paths_to_copy:
+            copyfile(src_file_path, os.path.join(dest_dir, new_file_name))
```
kevinrizza: I think this logic should be inside the `flatten` module. Take a look at the patterns of the other API methods, they do not have any abstractions related to the specific use of the method itself. If they have additional code, it is strictly related to converting input to something that is digestible by the underlying module. Here it seems like this is actual implementation detail of flattening - creating a directory and copying files. I would prefer something simpler like:

1. `create output directory if not exists`
2. `flatten.flatten(input_dir, output_dir)`

The actual implementation of the flattening should be left up to the `flatten` module.

operatorcourier/flatten.py
kevinrizza: +1

operatorcourier/flatten.py
```diff
@@ -0,0 +1,151 @@
...
+
+    file_paths_to_copy = []  # [ (SRC_FILE_PATH, NEW_FILE_NAME) ]
+
+    crd_dict = {}  # { CRD_NAME => (VERSION, CRD_PATH) }
+    csv_paths = []
+
+    for version_folder_name in folder_names:
+        parse_version_folder(source_dir, version_folder_name, csv_paths, crd_dict)
+
+    # 1. add package in source_dir
```
kevinrizza: I think these comments do not need numbering.

kevinrizza:
> > Also would be nice if dir structure is already flat to not raise an error but log only INFO message
> 
> To clarify this a little bit: we discussed with @ralphbean that in order to avoid confusion among OMPS users, the service should be able to handle both flattened and nested content, using the same endpoint. To be able to do this, OMPS either
> 
> * should be able to tell the difference between the two, and probably should rely on Operator Courier to get this information. This would require an `is_flat` or something similar API.
> 
> OR
> 
> * `flatten` should do nothing if called on a "flat" tree.

I would very strongly prefer the latter. If we aren't going to successfully create the new directory, we should just do nothing and log something indicating that happened.

operatorcourier/cli.py
jeremylinlin: @csomh This is strange... I am able to see the description on my side.

```
(operator-courier) 10:42 > operator-courier git:(add-flatten-command) ✗ operator-courier flatten -h
usage: operator-courier [-h] source_dir dest_dir

Given a directory with different versions of operator bundles (CRD, CSV,
package), this command extracts versioned CSVs and the latest version of each
CRD along with the package file and creates a new flat directory of yaml
files. See https://github.com/operator-framework/operator-registry#manifest-
format to find out more about how nested bundles should be structured.

positional arguments:
  source_dir  Path of the source directory that contains different versions of
              operator bundles (CRD, CSV, package)
  dest_dir    The new flat directory that contains extracted bundle files

optional arguments:
  -h, --help  show this help message and exit
```

operatorcourier/flatten.py
jeremylinlin: @csomh Without the `if` block, all CSVs with be appended with the version number. This part of the code is actually handling a corner case where two or more CSVs have the same file name. But in reality, we do not expect this to happen a lot (could you @kevinrizza confirm this?), and doing such transformation may usually end up duplicate version numbers in the file name.

operatorcourier/cli.py
csomh: @jeremylinlin, it's not about the description, it's the usage string. For `operator-courier flatten -h` my expectation would be

```
operator-courier flatten [-h] source_dir dest_dir
```

but it's:

```
operator-courier [-h] source_dir dest_dir
```

operatorcourier/flatten.py
jeremylinlin: @MartinBasti Do you mean that I need to recursively search for package files in the source_dir? I think we should only expect the package file to exist in the root directory and ignoring those in version folders.

operatorcourier/flatten.py
csomh: Are duplicate version numbers and issue?

kevinrizza: The filenames don't make a difference once we bundle, so I think it's fine if we don't have more complex logic here and instead just assume we want to rename the files.

operatorcourier/flatten.py
csomh: tbh, I don't see introducing type annotations as a priority, definitely not in this PR, and I would argue to remove them for now.

If you think type annotations  would improve the code quality, then better document this in the contribution doc, add a check for this in tox and update the whole project to meet the requirement.

operatorcourier/flatten.py
kevinrizza: If we assume the package is in the root directory then don't we not actually need to iterate over them?

operatorcourier/flatten.py
kevinrizza: +1

operatorcourier/cli.py
jeremylinlin: @csomh Gotcha. But it seems to be a general issue for all other subcommands. I fiddled around for some time and wasn't able to find a quick fix. Maybe we should open up an issue and make a separate PR for this? @kevinrizza

operatorcourier/flatten.py
jeremylinlin: @kevinrizza Yes iterating over the root directory is done [here](https://github.com/operator-framework/operator-courier/blob/2a9a333bcd/operatorcourier/flatten.py#L23) (but only for 1 level) , and the input of this helper is expected be only files coming from the root dir, as is indicated by its arg name `file_names_in_base_dir`.

MartinBasti: ok, if this is root dir I'm fine, I was afraid that it uses subdirectories.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,150 @@
...
+        yaml_type = identify.get_operator_artifact_type(file_content)
+
+        if yaml_type == 'ClusterServiceVersion':
+            contains_csv = True
+            csv_paths.append(item_path)
+        elif yaml_type == 'CustomResourceDefinition':
+            try:
+                crd_name = safe_load(file_content)['metadata']['name']
+            except (MarkedYAMLError, KeyError):
+                msg = "Courier requires valid input YAML files"
```
csomh: This message will be misleading in case of KeyErrors.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,150 @@
...
+
+    return packages[0]
+
+
+# parse all CSVs and ensure those with same names are handled
+def create_csv_dict(csv_paths: list) -> dict:
+    csv_dict = {}  # { CSV_FILE_NAME => [ (v1, csv_path_1), ... , (vn, csv_path_n) ] }
+    for csv_path in csv_paths:
+        version_folder_path, csv_file_name = os.path.split(csv_path)
+        version = version_folder_path.split('/')[-1]
```
csomh:
```suggestion
        version = os.path.basename(version_folder_path)
```

operatorcourier/flatten.py
jeremylinlin: @csomh  Actually I'm not sure if the `try-except` is really needed here. Because we have already called `identify.get_operator_artifact_type` before in the same method, and this call should include those checks already.

kevinrizza: @jeremylinlin all that method does is check the file type (either by checking the `kind` or explicitly checking the file's fields for a package file). I think we shouldn't make the assumption that the file is formatted correctly if we need the `metadata.name` field. Better to be safe and not run into an index error.

operatorcourier/cli.py
kevinrizza: It's a good point, but let's not check in more bad code. Let's do this one right and fix the others in a separate PR.

operatorcourier/flatten.py
SamiSousa: Is this matching the filename of the CSV or the CSV object name?
Also, should we be pushing our own opinions on the naming conventions or following existing conventions set by the operator-sdk? Or be flexible?
I ask because there are operators today that follow different conventions, ie. with and without the v, periods as separators, dashes, etc.

kevinrizza: Literally just the filename of the CSV. I think this problem you are describing @SamiSousa is orthogonal to the issue we are trying to solve. We are not suggesting or being opinionated at all about what you *should* name your files. We are just renaming them to avoid accidental collisions before we push. These files we create shouldn't be checked in or used for any other purpose, just used as an intermediary step at runtime.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,150 @@
...
+    version_folder_path = os.path.join(base_dir, version_folder_name)
+
+    for item in os.listdir(os.path.join(base_dir, version_folder_name)):
+        item_path = os.path.join(version_folder_path, item)
+
+        if not os.path.isfile(item_path):
+            logger.warning('Ignoring %s as it is not a regular file.', item)
+
+        with open(item_path, 'r') as f:
+            file_content = f.read()
```
SamiSousa: nit: Can the `file_content` variable be defined outside this `with` statement? I thought that `with` has it's own scope.

operatorcourier/flatten.py
SamiSousa: Could we include the KeyError in the log as well?

operatorcourier/flatten.py
```diff
@@ -0,0 +1,150 @@
...
+        elif yaml_type == 'CustomResourceDefinition':
+            try:
+                crd_name = safe_load(file_content)['metadata']['name']
+            except (MarkedYAMLError, KeyError):
+                msg = "Courier requires valid input YAML files"
+                logger.error(msg)
+                raise errors.OpCourierBadYaml(msg)
+            # create new CRD type entry if not found in dict
+            if crd_name not in crd_dict:
+                crd_dict[crd_name] = (version_folder_name, item_path)
```
SamiSousa: Should this use the crd's version or the folder (CSV) version? I'm confused about what the version means for the CRD, is it the CRD's API version or the CSV version that it comes with?

operatorcourier/api.py
```diff
@@ -136,3 +137,17 @@ def nest(source_dir, registry_dir):
 
     with TemporaryDirectory() as temp_dir:
         nest_bundles(yaml_files, registry_dir, temp_dir)
+
+
+def flatten(source_dir, dest_dir):
+    """
+    Given a directory containing different versions of operator bundles
+    (CRD, CSV, package) in separate version directories, this function
+    copies all files needed to the flattened directory
```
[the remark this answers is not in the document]
jeremylinlin: Sure!

operatorcourier/flatten.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # add all CRDs with the latest version
+    for _, crd_path in crd_dict.values():
+        file_paths_to_copy.append((crd_path, os.path.basename(crd_path)))
+
+    # add all CSVs
+    for csv_file_name, csv_entries in create_csv_dict(csv_paths).items():
+        for (version, csv_path) in csv_entries:
+            basename, ext = os.path.splitext(csv_file_name)
+            file_paths_to_copy.append((csv_path, f'{basename}-v{version}{ext}'))
```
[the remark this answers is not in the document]
jeremylinlin: > So here we're renaming the CSV files to avoid naming overlap, is that right?

Yes.

> Will the resulting file ext be of the format .clusterserviceversion.yaml?

It will append a version string before the extension. You can check the [test cases](https://github.com/operator-framework/operator-courier/blob/64095c9c50/tests/test_flatten.py#L18) to get a better sense.

operatorcourier/flatten.py
jeremylinlin: It seems that scope in Python is a little different than some other languages like Java. So I guess I do not have to declare it outside of the `with` block? https://stackoverflow.com/a/44437916/3949193

operatorcourier/flatten.py
jeremylinlin: @SamiSousa It's just the folder version. All versions in this flatten file refer to that.

operatorcourier/flatten.py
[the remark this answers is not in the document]
SamiSousa: Why add the version in two spots of the file name? Is it because the first version was part of the file name?
To keep with conventions for yaml naming, the file format for CSVs should look something like:
```
{basename}{separator}v{version}.clusterserviceversion.yaml
```
For example:
```
etcd-operator.v0.9.2.clusterserviceversion.yaml
```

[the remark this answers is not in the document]
kevinrizza: yeah but that's not guaranteed @SamiSousa , it's just a convention. we don't want this utility to fail just because someone broke a convention.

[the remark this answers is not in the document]
jeremylinlin: @SamiSousa See this [conversation](https://github.com/operator-framework/operator-courier/pull/85#discussion_r269600287) and [this](https://github.com/operator-framework/operator-courier/pull/85#discussion_r269476581).

operatorcourier/cli.py
jeremylinlin: @kevinrizza May I hard code the right usage for `flatten` for this time?
```python
    def flatten(self):
        parser = argparse.ArgumentParser(
            usage='operator-courier flatten [-h] source_dir dest_dir',
            description='Given a directory with different versions of '
                        'operator bundles (CRD, CSV, package), this command extracts '
                        'versioned CSVs and the latest version of each CRD along with '
                        'the package file and creates a new flat directory '
                        'of yaml files. See https://github.com/operator-framework/'
                        'operator-registry#manifest-format to find out more about '
                        'how nested bundles should be structured.')
```

Because after reading some [docs about subcommands](https://docs.python.org/3/library/argparse.html#sub-commands) and doing some experiments, it seems the current subcommands implementation is not the standard way of doing it. And it will be very hard to make the `flatten` help message work using the "right way" without breaking others.

I can also make a PR to fix the CLI subcommands before fixing this PR, but this may take a while to get through while the current PR is more urgent.

operatorcourier/flatten.py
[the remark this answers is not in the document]
SamiSousa: @kevinrizza Where would this fail based on a convention? It seems like we're renaming the files anyways, so in a way aren't we just setting our own convention?
It looks like we're pulling the `basename` from the original filename, so that would explain why the version pops up twice. I don't mind if we have the version back to back like so:
```
etcd-operator.v0.9.2-v0.9.2.clusterserviceversion.yaml
```
The reason I'm insisting is because other tools may naively be depending on this filenaming convention to discern which files are CSVs, which are CRDs, etc.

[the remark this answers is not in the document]
kevinrizza: Because there's nothing stopping anyone from naming their files whatever they want. This command literally just renames them so we don't get collisions. No other tool besides courier should ever touch these temp files -- if they do that is a mistake and should be corrected.

jeremylinlin:
@csomh No problem. Thank you so much for reviewing my PR and providing great feedback.

operatorcourier/flatten.py
```diff
@@ -0,0 +1,155 @@
...
+    contains_csv = False
+    version_folder_path = os.path.join(base_dir, version_folder_name)
+
+    for item in os.listdir(os.path.join(base_dir, version_folder_name)):
+        item_path = os.path.join(version_folder_path, item)
+
+        if not os.path.isfile(item_path):
+            logger.warning('Ignoring %s as it is not a regular file.', item)
+
+        with open(item_path, 'r') as f:
```
MartinBasti: I tested this PR and I got following error (because I had binary (zip) file in that directory by accident):

```
$ operator-courier flatten orig dest
'utf-8' codec can't decode byte 0xd8 in position 14: invalid continuation byte
```
There is no traceback, but I assume that this line the source of that error.

`push` and `verify` commands reads only files with `*.yaml` and `*.yml` extensions, IMHO `flatten` should work consistently with other commands.

kevinrizza: +1 This is valuable feedback. I think there should be two takeways from this: we certainly limit our parsing and copying in this command to just yaml files. And we should also explicitly log and return an error if the folder contains files which are not of that type.

jeremylinlin: @MartinBasti Thanks for the feedback! I've updated the code to handle that situation.

MartinBasti:
Actually coverage could be updated, https://coveralls.io/builds/22511114/source?filename=operatorcourier/flatten.py but it could be a new PR to not block this feature

operatorcourier/flatten.py
```diff
@@ -0,0 +1,170 @@
+import logging
+import os
+from yaml import safe_load, MarkedYAMLError
+from typing import Dict, Tuple
+from shutil import copyfile
+import semver
+import operatorcourier.identify as identify
```
MartinBasti: How about?
```python
from operatorcourier import identify
from operatorcourier import errors
```

jeremylinlin: Thanks @MartinBasti. Updated.

PR #89, opened 2019-03-29
CLI: Add --verbose mode option

MartinBasti:
* also show traceback in debug mode

### DIFF
--- a/operatorcourier/cli.py
+++ b/operatorcourier/cli.py
@@ -2,17 +2,20 @@ import argparse
 import pkg_resources
 import sys
 import logging
+import traceback
+
 from operatorcourier import api
 
 
 def main():
     """Generate the CLI bits
     """
-    logging.basicConfig()
     try:
         parser = _CliParser()
         parser.parse()
     except Exception as e:  # Exception raised to CLI should not be thrown to users,
+        logger = logging.getLogger(__name__)
+        logger.debug("Exit traceback: %s", traceback.format_exc())
         sys.exit(str(e))    # it should just be captured by logs
 
 
@@ -39,6 +42,10 @@ class _CliParser():
             '-v', '--version',
             help='Show the current version of operator-courier',
             action='version', version=__version__)
+        parser.add_argument(
+            '--verbose', dest='verbose',
+            help="Provide detailed logs",
+            action='store_true', default=False)
 
         subparsers = parser.add_subparsers(title='subcommands')
 
@@ -124,6 +131,10 @@ class _CliParser():
         flatten_parser.set_defaults(func=self.flatten)
 
         args = parser.parse_args()
+        logging.basicConfig(
+            level=logging.DEBUG if args.verbose else logging.WARNING
+        )
+
         args.func(args)
 
     def verify(self, args):


operatorcourier/cli.py
```diff
@@ -45,11 +48,18 @@ def parse(self):
         parser.add_argument('-v', '--version',
                             help='Show the current version of operator-courier',
                             action='version', version=__version__)
+        parser.add_argument('--log-level', dest='log_level',
+                            help="Set logging level (default: WARNING)",
+                            choices=['DEBUG', 'INFO', 'WARNING', 'ERROR',
+                                     'CRITICAL'],
+                            default='WARNING', type=str.upper)
 
         args = parser.parse_args(sys.argv[1:2])
```
MartinBasti: This PR is not working due this line. Because PR with argparse subparsers was rejected, I'm waiting for suggestion how to update this.

MartinBasti:
Reason for this PR was in https://github.com/operator-framework/operator-courier/pull/85#discussion_r270360877 as it was not nice to get which part of code failed from CLI output

operatorcourier/cli.py
kevinrizza: @MartinBasti 

Two comments here:

1. I'm still not sure I see why not using subparsers is preventing a top level flag like this. Why can't we just explicitly parse this command in this method? We aren't throwing anything away.

2. The PR that included a move to subparsers wasn't rejected because it used subparsers. It was rejected because the PR was attempting to address two unrelated things (adding subparsers and attempting to resolve this issue: https://github.com/operator-framework/operator-courier/issues/77). I just stated that moving to subparsers didn't fix that issue, and either the PR should have been updated to fix that issue and explain why adding subparser use was needed to resolve it OR stop referencing that issue and instead just add subparsers and in a detailed way explain why that was useful. Instead of following up on my point, the PR was just closed.

I am not against using subparsers, I just believe that git history should be concise and clean and that every PR should have a singular stated purpose. If you strongly believe that the best way to implement this feature is to include the use of subparsers then feel free to either reopen the other PR with what I was requesting, or feel free to include that change in this PR if you feel it is necessary for this change. But when you do, please clearly state why that is the case.

operatorcourier/cli.py
```diff
@@ -45,11 +48,18 @@ def parse(self):
         parser.add_argument('-v', '--version',
                             help='Show the current version of operator-courier',
                             action='version', version=__version__)
+        parser.add_argument('--log-level', dest='log_level',
+                            help="Set logging level (default: WARNING)",
+                            choices=['DEBUG', 'INFO', 'WARNING', 'ERROR',
```
kevinrizza: Do we really want to allow all of these options? Why not just allow DEBUG and INFO? By default we will already log the others.

MartinBasti: What is wrong about having all of them? when I don't care about WARNINGs I can enable only errors.

I can use `--verbose` or `--debug` to enable DEBUG level, if you insist

operatorcourier/cli.py
MartinBasti: 1. 
* main parser uses only first argument `sys.argv[1:2]`
* command parsers uses second+ argument for parsing `sys.argv[2:]`
* when I use `--log-level` on first position I got error that subcommand is required
* when I use subcommand on first position, log-level is just ignored
* because log-level is optional, I cannot hardcode `sys.argv[3:]`
* I'm not gonna copy that option to each command parser
* subparsers handles this even without thinking about that some `sys.argv[]` exists

2. I will reopen it then

operatorcourier/cli.py
kevinrizza: I would prefer `-v` and `--verbose` to enable DEBUG from the command line since this is a much more common convention.

MartinBasti: `-v` is already used

I will rework that to `--verbose`

MartinBasti:
Ready for review

MartinBasti:
Is there anything blocking this PR from merging?

kevinrizza:
> Is there anything blocking this PR from merging?

Was waiting on one more set of eyes. Thanks @jeremylinlin

PR #91, opened 2019-04-01
Add Optional Output File

kevinrizza:
Problem:
Currently, when running the verify or push commands there is no easy way
to actually inspect the bundle that is created -- it is a temp file that
is discarded when the command execution completes. Ideally, it would
be useful to optionally dump the bundle yaml to a specified file.

Solution:
Add optional `output` parameter that dumps the bundle to specified file.


tests/integration/test_verify.py
```diff
@@ -16,6 +17,24 @@ def test_verify_valid_sources(source_dir):
...
+def test_verify_with_output_file(source_dir, output):
+    process = subprocess.Popen(f'operator-courier verify {source_dir} --output={output}',
+                               shell=True,
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.STDOUT)
+    exit_code = process.wait()
+    assert exit_code == 0
+
+    fileCreated = os.path.isfile(output)
+    assert fileCreated
```
jeremylinlin: Should we update some variable naming styles according to [PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)? I've seen mixed use of camelCase and snake_case in our project.

kevinrizza: This is a great point, pr updated.

tests/integration/test_verify.py
```diff
@@ -16,6 +17,24 @@ def test_verify_valid_sources(source_dir):
...
+])
+def test_verify_with_output_file(source_dir, output):
+    process = subprocess.Popen(f'operator-courier verify {source_dir} --output={output}',
+                               shell=True,
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.STDOUT)
+    exit_code = process.wait()
+    assert exit_code == 0
+
+    fileCreated = os.path.isfile(output)
```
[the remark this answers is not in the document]
kevinrizza: I don't think it matters what the file is, we are creating it if it doesn't exist and overwriting it if it does. Whether they want to name their file with the correct file extension isn't really important, considering this is just a debug feature anyway.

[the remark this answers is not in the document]
jeremylinlin: Yeah but I'm not talking about the file name/extension. I'm thinking as integration tests, we should test the functionality/feature of the application. For this specific test case, It would be ideal if we can check the content of the output matches what we would expect. We may not need to go that far, but currently even an empty file an pass the check.

tests/integration/test_verify.py
```diff
@@ -16,6 +17,24 @@ def test_verify_valid_sources(source_dir):
     assert exit_code == 0
 
 
+@pytest.mark.parametrize('source_dir,output', [
+    ("tests/test_files/yaml_source_dir/valid_yamls_with_multiple_crds",
+     "tests/bundle.yaml"),
```
jeremylinlin: Maybe safer to put the output file in `/tmp`, just in case the test is aborted in the middle and the output file doesn't get deleted?

MartinBasti: +1, you can use pytest's `tmpdir` fixture or `tempfile` module

operatorcourier/api.py
```diff
@@ -22,7 +22,7 @@
 
 
 def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
-                     validation_output=None, repository=None):
```
MartinBasti: new options should be added at the end due BW compatibility.

operatorcourier/api.py
```diff
@@ -67,6 +67,9 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
 
     if valid:
         bundle = format_bundle(bundle)
+        if output is not None:
+            with open(output, 'w+') as f:
```
MartinBasti: why `w+`? `+` opens FD for both read and write

tests/integration/test_verify.py
```diff
@@ -16,6 +17,24 @@ def test_verify_valid_sources(source_dir):
...
+])
+def test_verify_with_output_file(source_dir, output):
+    process = subprocess.Popen(f'operator-courier verify {source_dir} --output={output}',
```
MartinBasti: Why do you need popen? Can you just use `api` function in unittest instead? (AFAIK there are no cli tests in operator-courier which should handle arg parsing and passing)

kevinrizza: This isn't a unit test. This is an integration test that ensures that, on the system, if you use the cli tool with this flag that a file actually gets created. We could also create a unit test that tests this code is executed, but that's not the purpose here.

kevinrizza:
@MartinBasti @jeremylinlin Rebase done and comments addressed. Please review when you have time. Thank you!

tests/integration/test_verify.py
```diff
@@ -16,6 +19,28 @@ def test_verify_valid_sources(source_dir):
...
+                                   shell=True,
+                                   stdout=subprocess.PIPE,
+                                   stderr=subprocess.STDOUT)
+        exit_code = process.wait()
+        assert exit_code == 0
+
+        file_created = os.path.isfile(output)
+        assert file_created
+
+        content_correct = filecmp.cmp(result_file, output)
```
MartinBasti: This naive comparison will be pretty hard to achieve as there is no order guaranteed in python dicts and depends on python version/implementation

kevinrizza: You're probably right but that is literally what we do right now for all of our tests. At the moment ordering seems consistent simply because these tests always pass (and because we are just copying blobs). But adding code that reads these files and checks their content is just more code that needs to be tested itself, which I am against -- that's why the original implementation I did doesn't bother to look at the file contents, just that the file itself was created. IMO if we wanted to solve that problem we would have to invest more significant effort in solving the problem in the general case, or just ignore it.

My preference is for now either remove this assertion, or keep it and remove it if we get in a case where the result is not guaranteed (due to some other change in the source).

MartinBasti: You can try to loaf that out file and expected file via yaml and compare those 2 objects at python level

kevinrizza: Yes but the minute we do that we need to codify what data is in the files in these tests. So if the logic about how these files are created is changed, then we need to change these tests. That is very brittle and IMO should be avoided.

MartinBasti: I don't understand what do you mean. If you change something in code you have to update `result/bundle.yaml` anyway. I still don't see where is the issue between
```python
assert load_file(output) == load_file('result/bundle.yaml')
```
and
```python
assert yaml.safe_load(load_file(output)) == yaml.safe_load(load_file('result/bundle.yaml'))
```

kevinrizza: Ah I see your point now. Yeah that's a great idea, it resolves the problem without worrying about the schema.

kevinrizza: Code updated, please review when you have time. Thanks Martin.

MartinBasti: Yeah, this didn't help. Order of processing of CRD files is different on ubuntu vs macos. And I forgot that it is yaml inside yaml, so without deeper deconstruction those data cannot be validated in this way.

Probably just load output file decode with yaml module and check id format is valid yaml and nonempty.

MartinBasti: For inspiration: https://github.com/release-engineering/operators-manifests-push-service/blob/master/tests/integration/utils.py#L26

kevinrizza: @MartinBasti But that test will break the minute the format changes. I specifically want to avoid encoding the data format of the bundle into the tests.

kevinrizza:
Latest set of tests all pass, please review when you get a chance. Thanks.

@MartinBasti @jeremylinlin

PR #93, opened 2019-04-03
flatten: copy files to dest_dir when structure is already flat

MartinBasti:
operator-courier flatten should copy source_dir to dest_dir when
structure is already flat.

It should support this worklow:
```
$ operator-courier flatten source_dir dest_dir && zip archive.zip
dest_dir/*
```

Currently it fails when dest_dir doesn't exist or constructs empty
archive.

### DIFF
--- a/operatorcourier/flatten.py
+++ b/operatorcourier/flatten.py
@@ -27,10 +27,14 @@ def get_flattened_files_info(source_dir: str) -> [(str, str)]:
     """
 
     # extract package file and version folders from source_dir
-    _, folder_names, file_names = next(os.walk(source_dir))
+    root, folder_names, file_names = next(os.walk(source_dir))
     if not folder_names:
         logger.info('The source directory is already flat.')
-        return []
+        # just return files from dir as they are already flat
+        return [
+            (os.path.join(root, name), name)
+            for name in file_names
+        ]
 
     file_paths_to_copy = []  # [ (SRC_FILE_PATH, NEW_FILE_NAME) ]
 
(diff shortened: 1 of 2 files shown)


operatorcourier/flatten.py
```diff
@@ -27,10 +27,14 @@ def get_flattened_files_info(source_dir: str) -> [(str, str)]:
...
-    _, folder_names, file_names = next(os.walk(source_dir))
+    root, folder_names, file_names = next(os.walk(source_dir))
     if not folder_names:
         logger.info('The source directory is already flat.')
-        return []
+        # just return files from dir as they are already flat
+        return [
+            (os.path.join(root, name), name)
+            for name in file_names
+        ]
```
kevinrizza: What about the possibility that the directory we are creating is in a bad state? What if the input dir is empty?

MartinBasti: If input directory is empty, flatten should fail. (I'll add check)

What do you mean by this  "directory we are creating is in a bad state"?

kevinrizza: Empty dir is an example of a bad state.

I'm saying that, in general, it's easier to reason about the positive case than the negative one. It's an easy statement to make that "we introspected and found that we have this nested directory structure, so let's do this flatten thing". It's much harder to say definitively that "we don't have this nested structure, so then it must be flat files" simply because it has less of a surface area. So the correct thing to do, imperatively, would instead have a statement like:

`if` this_is_a_flat_dir_with_the_right_files:
    do_the_flattening()`

Otherwise you have to deal with the edge case. In this case, I think part of that is an artifact of how this function is defined in general, so it's not easy for us to get around it. I'm okay with just checking that one edge case for now that we think of, but in general it would be better for us to refactor this whole method so that we could ask questions like that more easily.

PR #94, opened 2019-04-08
CliParser: Print help when no argument is passed

darkowlzz:
This fixes the issue in the CliParser that shows error when no argument
is passed by printing the help message and exiting.

```
$ operator-courier 
'Namespace' object has no attribute 'func'
```

### DIFF
--- a/operatorcourier/cli.py
+++ b/operatorcourier/cli.py
@@ -135,7 +135,12 @@ class _CliParser():
             level=logging.DEBUG if args.verbose else logging.WARNING
         )
 
-        args.func(args)
+        func = getattr(args, 'func', None)
+        if callable(func):
+            func(args)
+        else:
+            parser.print_help(sys.stderr)
+            sys.exit(2)
 
     def verify(self, args):
         """Run the verify command


operatorcourier/cli.py
```diff
@@ -123,6 +123,9 @@ def parse(self):
             'extracted bundle files')
         flatten_parser.set_defaults(func=self.flatten)
 
+        if len(sys.argv) == 1:
+            parser.print_help(sys.stderr)
+            sys.exit(1)
```
MartinBasti: exit code 2 is usually used for argument parsing errors

operatorcourier/cli.py
```diff
@@ -123,6 +123,9 @@ def parse(self):
             'extracted bundle files')
         flatten_parser.set_defaults(func=self.flatten)
 
+        if len(sys.argv) == 1:
```
MartinBasti: this check should be added after `parser.parse_args()` because `operator-courier --version` is valid

Also I'd go with checking presence of `arg.func` instead testing number of args.

darkowlzz: Thanks. I've added a check for the presence of `func` and if it's callable.

darkowlzz:
Updated with the requested changes.

darkowlzz:
Rebased.

PR #98, opened 2019-04-11
use OS specific path separator

darkowlzz:
Replaces all the path creation lines with os.path.join to use os.sep.

### DIFF
--- a/operatorcourier/api.py
+++ b/operatorcourier/api.py
@@ -51,7 +51,7 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
     if source_dir is not None:
         for filename in os.listdir(source_dir):
             if filename.endswith(".yaml") or filename.endswith(".yml"):
-                with open(source_dir + "/" + filename) as f:
+                with open(os.path.join(source_dir, filename)) as f:
                     yaml_files.append(f.read())
     elif yamls is not None:
         yaml_files = yamls
@@ -111,7 +111,7 @@ def build_verify_and_push(namespace, repository, revision, token,
                               validation_output=validation_output, output=output)
 
     with TemporaryDirectory() as temp_dir:
-        with open('%s/bundle.yaml' % temp_dir, 'w') as outfile:
+        with open(os.path.join(temp_dir, 'bundle.yaml'), 'w') as outfile:
             yaml.dump(bundle, outfile, default_flow_style=False)
             outfile.flush()
 
@@ -135,7 +135,7 @@ def nest(source_dir, registry_dir):
     if source_dir is not None:
         for filename in os.listdir(source_dir):
             if filename.endswith(".yaml") or filename.endswith(".yml"):
-                with open(source_dir + "/" + filename) as f:
+                with open(os.path.join(source_dir, filename)) as f:
                     yaml_files.append(f.read())
 
     with TemporaryDirectory() as temp_dir:
--- a/operatorcourier/nest.py
+++ b/operatorcourier/nest.py
@@ -38,7 +38,8 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
 
     # write the package file
     package_name = package["packageName"]
-    with open('%s/%s.package.yaml' % (temp_registry_dir, package_name), 'w') as outfile:
+    packagefile_name = os.path.join(temp_registry_dir, '%s.package.yaml' % package_name)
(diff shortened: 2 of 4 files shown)


operatorcourier/validate.py
```diff
@@ -85,7 +85,7 @@ def _bundle_validation(self, bundle, repository=None):
                 validationDict[self.pkgsKey] = False
 
         valid = True
-        for key, value in validationDict.items():
+        for _, value in validationDict.items():
```
kevinrizza: While this seems like a valid thing to do, it seems unrelated to the purpose of this pr. Can you please remove it and create a separate pr for it? Or at least a separate commit and explain why this change is being made in a separate commit message?

darkowlzz: Sure, thanks.

operatorcourier/api.py
```diff
@@ -111,7 +111,7 @@ def build_verify_and_push(namespace, repository, revision, token,
                               validation_output=validation_output, output=output)
 
     with TemporaryDirectory() as temp_dir:
-        with open('%s/bundle.yaml' % temp_dir, 'w') as outfile:
+        with open(os.path.join('%s', 'bundle.yaml') % temp_dir, 'w') as outfile:
```
jeremylinlin:
```suggestion
        with open(os.path.join(temp_dir, 'bundle.yaml'), 'w') as outfile:
```

operatorcourier/push.py
```diff
@@ -32,7 +33,7 @@ def push(self, bundle_dir, namespace, repository, release, auth_token):
 
     def _create_base64_bundle(self, bundle_dir, repository):
         with TemporaryDirectory() as temp_dir:
-            tarfile_name = "%s/%s.tar.gz" % (temp_dir, repository)
+            tarfile_name = os.path.join("%s", "%s.tar.gz") % (temp_dir, repository)
```
jeremylinlin:
```suggestion
            tarfile_name = os.path.join(temp_dir, "%s.tar.gz") % repository
```

operatorcourier/nest.py
```diff
@@ -38,7 +38,9 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
 
     # write the package file
     package_name = package["packageName"]
-    with open('%s/%s.package.yaml' % (temp_registry_dir, package_name), 'w') as outfile:
+    packagefile_name = os.path.join('%s', '%s.package.yaml') % (
```
jeremylinlin:
```suggestion
    packagefile_name = os.path.join(temp_registry_dir, '%s.package.yaml') % package_name
```

operatorcourier/nest.py
```diff
@@ -62,7 +64,8 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
             crd_name = csv_crd["name"]
             if crd_name in crds:
                 crd = crds[crd_name]
-                with open('%s/%s.crd.yaml' % (csv_folder, crd_name), 'w') as outfile:
+                crdfile_name = os.path.join('%s', '%s.crd.yaml') % (csv_folder, crd_name)
```
jeremylinlin:
```suggestion
                crdfile_name = os.path.join(csv_folder, '%s.crd.yaml') % crd_name
```

jeremylinlin:
Looks good to me overall. Just some formatting suggestions that may make code easier to read.

operatorcourier/nest.py
MartinBasti: nope, this is correct
```python
os.path.join(csv_folder, '%s.crd.yaml' % crd_name')
```

operatorcourier/nest.py
MartinBasti:
```python
os.path.join(temp_registry_dir, '%s.package.yaml' % package_name)
```

operatorcourier/nest.py
MartinBasti: resp. both are correct, but way I mentioned seems more clear to me and less error prone

operatorcourier/validate.py
MartinBasti:
```python
for value in validationDict.values():
```

operatorcourier/nest.py
jeremylinlin: @MartinBasti Thanks for the suggestion!

darkowlzz:
Thanks for all the suggestions. I've updated the two commits with the suggested changes.

PR #102, opened 2019-04-11
Support nested directory layout

jeremylinlin:
Today the appregistry-server in operator-registry assumes that the bundles are stored in a flat format in the Quay appregistry namespace. We are migrating to allowing the namespace to have images in the nested format as supported natively by operator-registry.

### DIFF
--- a/operatorcourier/api.py
+++ b/operatorcourier/api.py
@@ -8,15 +8,12 @@ when using the operator-courier package
 import os
 import logging
 from tempfile import TemporaryDirectory
+from distutils.dir_util import copy_tree
 import yaml
-import json
-from operatorcourier.build import BuildCmd
-from operatorcourier.validate import ValidateCmd
+from operatorcourier.verified_manifest import VerifiedManifest
 from operatorcourier.push import PushCmd
-from operatorcourier.format import format_bundle
 from operatorcourier.nest import nest_bundles
 from operatorcourier.flatten import flatten_bundles
-from operatorcourier.errors import OpCourierBadBundle
 
 logger = logging.getLogger(__name__)
 
@@ -26,7 +23,9 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
     """Build and verify constructs an operator bundle from
     a set of files and then verifies it for usefulness and accuracy.
 
-    It returns the bundle as a string.
+    This API verifies the input source_dir or yamls and returns a verification
+    object which contains a nested boolean that indicates if the source_dir
+    is nested or not, and a bundle dictionary if the source_dir is in the flat structure.
 
     :param source_dir: Path to local directory of yaml files to be read.
     :param yamls: List of yaml strings to create bundle with
@@ -41,39 +40,17 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
     :raises OpCourierBadBundle: When the resulting bundle fails validation
     """
 
-    if source_dir is not None and yamls is not None:
-        logger.error("Both source_dir and yamls cannot be defined.")
-        raise TypeError(
-            "Both source_dir and yamls cannot be specified on function call.")
+    if source_dir and yamls:
+        msg = 'source_dir and yamls cannot both be specified.'
+        logger.error(msg)
+        raise TypeError(msg)
 
-    yaml_files = []
+    verified_manifest = VerifiedManifest(source_dir, yamls, ui_validate_io, repository)
 
-    if source_dir is not None:
-        for filename in os.listdir(source_dir):
-            if filename.endswith(".yaml") or filename.endswith(".yml"):
-                with open(os.path.join(source_dir, filename)) as f:
-                    yaml_files.append(f.read())
-    elif yamls is not None:
-        yaml_files = yamls
+    if validation_output:
+        verified_manifest.write_validation_to_file(validation_output)
 
-    bundle = BuildCmd().build_bundle(yaml_files)
-
-    valid, validation_results_dict = ValidateCmd(ui_validate_io).validate(bundle,
-                                                                          repository)
-
-    if validation_output is not None:
-        with open(validation_output, 'w') as f:
-            f.write(json.dumps(validation_results_dict) + "\n")
-
-    if valid:
-        bundle = format_bundle(bundle)
-        return bundle
-    else:
-        logger.error("Bundle failed validation.")
-        raise OpCourierBadBundle(
-            "Resulting bundle is invalid, input yaml is improperly defined.",
-            validation_info=validation_results_dict
-        )
+    return verified_manifest
 
 
 def build_verify_and_push(namespace, repository, revision, token,
@@ -103,16 +80,18 @@ def build_verify_and_push(namespace, repository, revision, token,
     :raises OpCourierQuayErrorResponse: When Quay responds with an error
     :raises OpCourierQuayError: When the request fails in an unexpected way
     """
+    verified_manifest = build_and_verify(source_dir, yamls, repository=repository,
+                                         validation_output=validation_output)
 
-    bundle = build_and_verify(source_dir, yamls, repository=repository,
-                              validation_output=validation_output)
-
-    with TemporaryDirectory() as temp_dir:
-        with open(os.path.join(temp_dir, 'bundle.yaml'), 'w') as outfile:
-            yaml.dump(bundle, outfile, default_flow_style=False)
-            outfile.flush()
-
-        PushCmd().push(temp_dir, namespace, repository, revision, token)
+    if not verified_manifest.nested:
+        with TemporaryDirectory() as temp_dir:
+            with open(os.path.join(temp_dir, 'bundle.yaml'), 'w') as outfile:
+                yaml.dump(verified_manifest.bundle, outfile, default_flow_style=False)
+            PushCmd().push(temp_dir, namespace, repository, revision, token)
+    else:
+        with TemporaryDirectory() as temp_dir:
+            copy_tree(source_dir, temp_dir)
+            PushCmd().push(temp_dir, namespace, repository, revision, token)
 
 
 def nest(source_dir, registry_dir):
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -19,8 +19,9 @@ class ValidateCmd():
     csvKey = "clusterServiceVersions"
     pkgsKey = "packages"
 
-    def __init__(self, ui_validate_io=False):
+    def __init__(self, ui_validate_io=False, nested=False):
         self.ui_validate_io = ui_validate_io
+        self.nested = nested
         self.validation_json = dict(
             warnings=[],
             errors=[],
@@ -52,6 +53,7 @@ class ValidateCmd():
 
         :param bundle: Dictionary of bundle value
         :param repository: Repository name for the application
+        :param nested: The input source is in nested structure or not
         """
         logger.info("Validating bundle.")
 
@@ -339,12 +341,11 @@ class ValidateCmd():
 
                     if "currentCSV" not in channel:
                         self._log_error("package channel.currentCSV not defined.")
-                    else:
-                        if channel["currentCSV"] not in csvNames:
-                            self._log_error("channel.currentCSV %s is not "
-                                            "included in list of csvs",
-                                            channel["currentCSV"])
-                            valid = False
+                    elif not self.nested and channel["currentCSV"] not in csvNames:
+                        self._log_error("channel.currentCSV %s is not "
+                                        "included in list of csvs",
+                                        channel["currentCSV"])
+                        valid = False
 
         else:
             self._log_error("package channels not defined.")
--- a/operatorcourier/verified_manifest.py
+++ b/operatorcourier/verified_manifest.py
@@ -1,0 +1,154 @@
+import os
+import logging
+import semver
+import json
+from operatorcourier.build import BuildCmd
+from operatorcourier.validate import ValidateCmd
+from operatorcourier.errors import OpCourierBadBundle
+from operatorcourier.format import format_bundle
+from operatorcourier import identify
+
+logger = logging.getLogger(__name__)
+FLAT_KEY = '__flat__'
+
+
+class VerifiedManifest:
+    @property
+    def bundle(self):
+        if self.nested:
+            raise AttributeError('VerifiedManifest does not have the bundle property '
+                                 'in nested cases.')
+        return format_bundle(self.bundle_dict)
+
+    def __init__(self, source_dir, yamls, ui_validate_io, repository):
+        if yamls:
+            manifests = dict(FLAT_KEY=yamls)
+        else:
+            manifests = self.get_manifests_info(source_dir)
+
+        self.nested = len(manifests) > 1
+        self.bundle_dict = None
+        self.__validation_dict = \
+            self.get_validation_dict_from_manifests(manifests, ui_validate_io, repository)
+
+    def get_manifest_files_content(self, file_paths):
+        """
+        Given a list of file paths, the function returns a list of strings containing
+        file content in all yaml files.
+        :param file_paths: a list of file paths
+        :return: a list of strings containing file content in all yaml files
+        """
+        manifests_content = []
+        for file_path in file_paths:
+            if file_path.endswith(".yaml") or file_path.endswith(".yml"):
+                with open(file_path) as f:
+                    manifests_content.append(f.read())
+        return manifests_content
+
+    def is_dir_name_semver(self, dir_name):
+        """
+        Checks if the dir_name is a semver or not.
+        :param dir_name: the name of the directory
+        :return: True if the dir_name is a semver, False otherwise
+        """
+        try:
+            semver.parse(dir_name)
+        except ValueError:
+            logging.warning('Ignoring directory "%s" as it is not in semver format.')
+            return False
+        return True
+
+    def get_manifests_info(self, source_dir):
+        """
+        Given a source directory OR a list of yaml files. The function returns a dict
+        containing all operator bundle file information grouped by version. Note that only
+        one of source_dir or yamls can be specified.
+
+        :param source_dir: Path to local directory of operator bundles, which can be
+                           either flat or nested
+        :param yamls: A list of yaml strings to create bundle with
+        :return: A dictionary object where the key is the semantic version of each bundle,
+                 and the value is a list of yaml strings of operator bundle files.
+                 FLAT_KEY is used as key if the directory structure is flat
+        """
+        # VERSION => manifest_files_content
+        # FLAT_KEY is used as key to indicate the flat directory structure
+        manifests = {}
+
+        root_path, dir_names, root_dir_files = next(os.walk(source_dir))
+        # removing dirs whose name are not semver
+        version_dirs = list(filter(lambda x: self.is_dir_name_semver(x), dir_names))
+        # flat directory
+        if not version_dirs:
+            manifests[FLAT_KEY] = self.get_manifest_files_content(
+                [os.path.join(root_path, file) for file in root_dir_files])
+        # nested
+        else:
+            # add all manifest files from each version folder to manifests dict
+            for version_dir in version_dirs:
+                version_dir_path = os.path.join(root_path, version_dir)
+                _, _, version_dir_files = next(os.walk(version_dir_path))
+                file_paths = [os.path.join(version_dir_path, file)
+                              for file in version_dir_files]
+                manifests[version_dir] = self.get_manifest_files_content(file_paths)
+            # get the package file from root dir and add to each version of manifest
+            package_content = None
+            for root_dir_file in root_dir_files:
+                with open(os.path.join(root_path, root_dir_file), 'r') as f:
+                    file_content = f.read()
+                if identify.get_operator_artifact_type(file_content) == 'Package':
+                    # ensure only 1 package is found in root directory
+                    if package_content:
+                        msg = 'There should be only 1 package file defined ' \
+                              'in the source directory.'
+                        logging.error(msg)
+                        raise OpCourierBadBundle(msg, {})
+                    package_content = file_content
+            if not package_content:
+                msg = 'No package file exists in the nested bundle.'
+                logging.error(msg)
+                raise OpCourierBadBundle(msg, {})
+            for version in manifests:
+                manifests[version].append(package_content)
+        return manifests
+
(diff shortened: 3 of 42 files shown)


operatorcourier/api.py
```diff
@@ -26,13 +24,15 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
...
-    It returns the bundle as a string.
+    It returns the bundle as a string, or None if the source_dir is nested.
 
     :param source_dir: Path to local directory of yaml files to be read.
     :param yamls: List of yaml strings to create bundle with
     :param ui_validate_io: Optional flag to test operatorhub.io specific validation
     :param validation_output: Path to optional output file for validation logs
     :param repository: Repository name for the application
+    :param output: The path to the operator bundle yaml output. This option is effective
+                   only if the source_dir is not in the nested structure
```
MartinBasti: Why? What produces build for nested structure?

operatorcourier/api.py
```diff
@@ -41,42 +41,29 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
     :raises OpCourierBadBundle: When the resulting bundle fails validation
     """
 
-    if source_dir is not None and yamls is not None:
-        logger.error("Both source_dir and yamls cannot be defined.")
-        raise TypeError(
-            "Both source_dir and yamls cannot be specified on function call.")
+    manifests = get_manifests_info(source_dir, yamls)
+    validation_dict, bundle_dict \
+        = get_validation_dict_from_manifests(manifests, ui_validate_io, repository)
```
MartinBasti: function is named `get_validation_dict_*` but returns `bundle_dict` too. It is quite messy IMHO

operatorcourier/cli.py
```diff
@@ -67,7 +67,8 @@ def parse(self):
         verify_parser.add_argument(
             '--validation-output',
             dest='validation_output',
-            help='A file to write validation warnings and errors to in JSON format')
+            help='A file to write validation warnings and errors to in JSON format.'
+                 'This option only works with the flat directory structure.')
```
MartinBasti: Why validation input cannot be saved to file?

operatorcourier/cli.py
```diff
@@ -98,7 +99,8 @@ def parse(self):
         push_parser.add_argument(
             '--validation-output',
             dest='validation_output',
-            help='A file to write validation warnings and errors to in JSON format')
+            help='A file to write validation warnings and errors to in JSON format.'
+                 'This option only works with the flat directory structure.')
```
MartinBasti: Why?

operatorcourier/verify.py
```diff
@@ -0,0 +1,131 @@
...
+    one of source_dir or yamls can be specified.
+
+    :param source_dir: Path to local directory of operator bundles, which can be either
+                       flat or nested
+    :param yamls: A list of yaml strings to create bundle with
+    :return: A dictionary object where the key is the the semantic version of each bundle,
+             and the value is a list of yaml strings of operator bundle files.
+             If the directory is flat, the key of yaml strings will be an empty string.
+    """
+    def get_manifest_files_content(file_paths: List) -> List:
```
MartinBasti: Could you please extract `get_manifest_files_content` and `is_dir_name_semver` from this function to module? I don't see reason why it needs closure.

operatorcourier/verify.py
```diff
@@ -0,0 +1,131 @@
...
+            return False
+        return True
+
+    if source_dir and yamls:
+        msg = 'source_dir and yamls cannot both be specified.'
+        logger.error(msg)
+        raise TypeError(msg)
+
+    # VERSION => manifest_files_content
+    # '' is used as key to indicate the flat directory structure
```
MartinBasti: use constant for this, like `FLAT_KEY = '__flat__'`

operatorcourier/verify.py
```diff
@@ -0,0 +1,131 @@
...
+                    file_content = f.read()
+                if identify.get_operator_artifact_type(file_content) == 'Package':
+                    # ensure only 1 package is found in root directory
+                    if packages_content:
+                        msg = 'There should be only 1 package file defined ' \
+                              'in the source directory'
+                        logging.error(msg)
+                        raise OpCourierBadBundle(msg, {})
+                    packages_content.append(file_content)
+            [manifests[version].append(packages_content[0]) for version in manifests]
```
MartinBasti: Please use standard for cycle instead creation of list

operatorcourier/verify.py
```diff
@@ -0,0 +1,131 @@
...
+    # validate on all bundles files and combine log messages
+    validation_dict = ValidateCmd(ui_validate_io).validation_json
+    for version, manifest_files_content in manifests.items():
+        bundle_dict = BuildCmd().build_bundle(manifest_files_content)
+        if version:
+            logging.info("Parsing version: %s", version)
+        _, validation_dict_temp = ValidateCmd(ui_validate_io).validate(bundle_dict,
+                                                                       repository, nested)
+        for log_level, msg_list in validation_dict_temp.items():
+            validation_dict[log_level].extend(validation_dict_temp[log_level])
```
MartinBasti: `.extend(msg_list)` ?

operatorcourier/cli.py
kevinrizza: Because the nested directory isn't generating a bundle at all. It actually zips up the directory itself, intact. We aren't generating yaml in this case, so there's no file to be generated.

operatorcourier/api.py
kevinrizza: Nothing. This IMO needs to be refactored more significantly because returning a bundle doesn't really make sense in this case anymore.

operatorcourier/api.py
```diff
@@ -41,42 +41,29 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
...
             with open(output, 'w') as f:
                 yaml.dump(bundle, f, default_flow_style=False)
-        return bundle
-    else:
-        logger.error("Bundle failed validation.")
-        raise OpCourierBadBundle(
-            "Resulting bundle is invalid, input yaml is improperly defined.",
-            validation_info=validation_results_dict
-        )
+    return bundle
```
kevinrizza: We need to what we are returning here and why. IMO this change needs to either be a breaking change, or the API needs to be refactored such that there is a separate method for the API in the nested case. My personal preference would be to refactor this method to return something other than `bundle` and break the API, since it is cleaner for courier to determine if the folder is nested or not. But @MartinBasti @jeremylinlin @awgreene please suggest alternatives here.

Either way, I don't think this should be as straightforward as it is defined here. This is very messy and will get confusing quickly.

operatorcourier/api.py
jeremylinlin: @MartinBasti Thank you for the comment! I have updated and set `bundle_dict` as an instance variable instead.

jeremylinlin:
@MartinBasti @csomh I have updated the PR with some refactoring after some discussion with @kevinrizza . A `Verification` object is created every time a `source_dir` is parsed and validated.

Now the `build_and_verify` API returns an instance of `verification` obj that contains
 - `validation_dict` where all errors and warnings are stored
 - `nested` boolean which indicates if `source_dir` is nested or flat, and
 - a `bundle` object which is set when `source_dir` is nested

Please provide some feedback when you have a chance. Thanks!

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+logger = logging.getLogger(__name__)
+FLAT_KEY = '__flat__'
+
+
+class Verification:
+    def __init__(self, source_dir, yamls, ui_validate_io, repository):
+        manifests = self.get_manifests_info(source_dir, yamls)
+        self.nested = len(manifests) > 1
+        self.bundle_dict = None
+        self.validation_dict = \
```
kevinrizza: Why do we need to pass the validation_dict around at all?

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+from operatorcourier.validate import ValidateCmd
+from operatorcourier.errors import OpCourierBadBundle
+from operatorcourier.format import format_bundle
+from operatorcourier import identify
+
+logger = logging.getLogger(__name__)
+FLAT_KEY = '__flat__'
+
+
+class Verification:
```
kevinrizza: Verification seems like the wrong name -- we already have a thing called `Validate` which seems very similar (although they serve very different meanings). Also, this thing does not actually _do_ any of the verification. Maybe this is a `Manifest` or `Verified_Manifest` [(from operator-registry)](https://github.com/operator-framework/operator-registry#manifest-format)

operatorcourier/validate.py
```diff
@@ -91,7 +95,7 @@ def _bundle_validation(self, bundle, repository=None):
 
         return valid
 
-    def _crd_validation(self, bundleData):
+    def _crd_validation(self, bundleData, nested):
```
kevinrizza: Why are we passing this variable along to each of these methods and then not using it for anything?

operatorcourier/validate.py
```diff
@@ -339,24 +343,23 @@ def _pkgs_validation(self, bundleData):
 
                     if "currentCSV" not in channel:
                         self._log_error("package channel.currentCSV not defined.")
-                    else:
-                        if channel["currentCSV"] not in csvNames:
-                            self._log_error("channel.currentCSV %s is not "
-                                            "included in list of csvs",
-                                            channel["currentCSV"])
-                            valid = False
+                    elif not nested and channel["currentCSV"] not in csvNames:
```
kevinrizza: This seems unnecessary and adds a lot of indirection to the method definitions. It's not really in the scope of the PR to add this check, right? Why add this?

operatorcourier/api.py
```diff
@@ -106,16 +83,15 @@ def build_verify_and_push(namespace, repository, revision, token,
...
-        PushCmd().push(temp_dir, namespace, repository, revision, token)
+    verification = build_and_verify(source_dir, yamls, repository=repository,
+                                    validation_output=validation_output, output=output)
+    if not verification.nested:
+        with TemporaryDirectory() as temp_dir:
+            with open(os.path.join(temp_dir, 'bundle.yaml'), 'w') as outfile:
+                yaml.dump(verification.bundle, outfile, default_flow_style=False)
+            PushCmd().push(temp_dir, namespace, repository, revision, token)
+    else:
+        PushCmd().push(source_dir, namespace, repository, revision, token)
```
kevinrizza: We should be careful here. My recommendation would be to copy the files into a temp directory and then create the tarball from there. It seems dangerous to use the source files directly as input considering we are making os calls other than read.

operatorcourier/api.py
```diff
@@ -41,42 +40,20 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
...
-                with open(os.path.join(source_dir, filename)) as f:
-                    yaml_files.append(f.read())
-    elif yamls is not None:
-        yaml_files = yamls
-
-    bundle = BuildCmd().build_bundle(yaml_files)
-
-    valid, validation_results_dict = ValidateCmd(ui_validate_io).validate(bundle,
-                                                                          repository)
+    verification = Verification(source_dir, yamls, ui_validate_io, repository)
```
kevinrizza: Why are we removing the check to see if `source_dir` and `yamls` are both defined?

kevinrizza: Ah I see, it moved to the Verification class. But doesn't it seem like encapsulating the mechanical part should stay in the API method (creating folders, checking what type of input, etc. ?)

operatorcourier/verify.py
jeremylinlin: @kevinrizza Could you please elaborate on your concern?

operatorcourier/validate.py
jeremylinlin: Because we need `nested` value for `_pkgs_validation`, which is one of the [`validators`](https://github.com/jeremylinlin/operator-courier/blob/support-nested-directory-layout/operatorcourier/validate.py#L362). By changing the function signature of this, we also need to update the other ones, otherwise errors will be thrown.

operatorcourier/validate.py
jeremylinlin: @kevinrizza Note that I only added `not nested` to the if condition. We need to remove this check for nested cases because `currentCSV` field in the package file can only match 1 CSV version. This will not work in nested cases where there are multiple CSVs. I tried to pass `nested` field inside also because of this.

operatorcourier/api.py
jeremylinlin: @kevinrizza The check is moved to [here](https://github.com/jeremylinlin/operator-courier/blob/support-nested-directory-layout/operatorcourier/verify.py#L65). But all it does is just checking inputs. I think it makes sense to stay in that method. Even if we move it to the API method, we will still need this check in `get_manifests_info`, right?

operatorcourier/cli.py
MartinBasti: But this is validation output, there is no validation for nested structure?

operatorcourier/api.py
MartinBasti: I'm also curious how this will work when yamls are defined instead of source_dir. (I don't need yamls param for my use cases, so it can be dropped if nobody else is using that)

MartinBasti: Got it, yamls are supported only as flat

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+                    if identify.get_operator_artifact_type(file_content) == 'Package':
+                        # ensure only 1 package is found in root directory
+                        if packages_content:
+                            msg = 'There should be only 1 package file defined ' \
+                                  'in the source directory'
+                            logging.error(msg)
+                            raise OpCourierBadBundle(msg, {})
+                        packages_content.append(file_content)
+                for version in manifests:
+                    manifests[version].append(packages_content[0])
```
MartinBasti: What if packages_content is empty?

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+                    file_paths = [os.path.join(version_dir_path, file)
+                                  for file in version_dir_files]
+                    manifests[version_dir] = self.get_manifest_files_content(file_paths)
+                # get the package file from root dir and add to each version of manifest
+                for root_dir_file in root_dir_files:
+                    with open(os.path.join(root_path, root_dir_file), 'r') as f:
+                        file_content = f.read()
+                    if identify.get_operator_artifact_type(file_content) == 'Package':
+                        # ensure only 1 package is found in root directory
+                        if packages_content:
```
MartinBasti: Why is this list when only one package is allowed?

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+        of the manifest (or FLAT_KEY if the manifest files are not grouped by version)
+        :param ui_validate_io: the ui_validate_io flag specified from CLI
+        :param repository: the repository value specified from CLI
+        :return: a dict containing validation info (warnings/errors).
+        """
+        # validate on all bundles files and combine log messages
+        validation_dict = ValidateCmd(ui_validate_io).validation_json
+        for version, manifest_files_content in manifests.items():
+            bundle_dict = BuildCmd().build_bundle(manifest_files_content)
+            if version:
```
MartinBasti: Is possible that version could be evaluated as False?

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+                validation_dict[log_level].extend(msg_list)
+
+        if validation_dict['errors']:
+            logger.error("Bundle failed validation.")
+            raise OpCourierBadBundle(
+                "Resulting bundle is invalid, input yaml is improperly defined.",
+                validation_info=validation_dict)
+
+        if not self.nested:
+            self.bundle_dict = bundle_dict
```
MartinBasti: `bundle_dict` may be used before assigned

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+
+class Verification:
+    def __init__(self, source_dir, yamls, ui_validate_io, repository):
+        manifests = self.get_manifests_info(source_dir, yamls)
+        self.nested = len(manifests) > 1
+        self.bundle_dict = None
+        self.validation_dict = \
+            self.get_validation_dict_from_manifests(manifests, ui_validate_io, repository)
+        if not self.nested:
+            self.bundle = format_bundle(self.bundle_dict)
```
MartinBasti: `bundle` could be a property with content generated on demand/raising exception for nested

operatorcourier/api.py
kevinrizza: Not if you have separate methods under the hood that do the validation. Why not have separate constructors depending on the type of input you get? Abstracting the mechanics of what to do with the input and calling the right functions based on that seems like the only real job of the API method.

operatorcourier/cli.py
kevinrizza: Sorry I think I misunderstood which output file we were referring to here. It seems like this should work on both.

operatorcourier/validate.py
kevinrizza: This is definitely a code smell. Passing these functions along without a way to abstract the function header seems like bad design. If we genuinely have different validation depending on whether or not a package is nested, we should have some way of abstracting that at the top level -- perhaps by having different root methods that call the same underlying validation methods on certain subfields. Otherwise this will become extremely difficult to maintain as we get new validation needs.

operatorcourier/validate.py
kevinrizza: So if that's the case I think this needs more careful consideration -- perhaps a method that does nested verification on the package file separate from the non nested case. See my comments above.

operatorcourier/verify.py
kevinrizza: We just use that function to write to a file. Why do we need to return it? Can't we just write to the file in the `build_and_verify()` method? What does the `push()` method need it for at all?

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+        :return: True if the dir_name is a semver, False otherwise
+        """
+        try:
+            semver.parse(dir_name)
+        except ValueError:
+            logging.warning('Ignoring directory "%s" as it is not in semver format.')
+            return False
+        return True
+
+    def get_manifests_info(self, source_dir: str, yamls: List) -> Dict:
```
kevinrizza: Same comment as above about return value.

operatorcourier/verify.py
jeremylinlin: @kevinrizza  I'm not sure if I follow your point. Do you want me to move all the code in `get_validation_dict_from_manifests()` method to `build_and_verify()`?

operatorcourier/verify.py
```diff
@@ -0,0 +1,143 @@
...
+    def __init__(self, source_dir, yamls, ui_validate_io, repository):
+        manifests = self.get_manifests_info(source_dir, yamls)
+        self.nested = len(manifests) > 1
+        self.bundle_dict = None
+        self.validation_dict = \
+            self.get_validation_dict_from_manifests(manifests, ui_validate_io, repository)
+        if not self.nested:
+            self.bundle = format_bundle(self.bundle_dict)
+
+    def get_manifest_files_content(self, file_paths: List) -> List:
```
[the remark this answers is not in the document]
jeremylinlin: @kevinrizza Since Python is dynamically typed, I just think that when it's not obvious what some params / return values' type is, it may help to explicitly add type hints.

operatorcourier/verify.py
jeremylinlin: @MartinBasti So what's the best way to prevent this? Maybe initialize `self.bundle` as `None`?

jeremylinlin:
@MartinBasti @kevinrizza Thank you for the comments! I have addressed most of them and have just a few that needs more clarifications. Please take a look when you have a chance.

operatorcourier/verify.py
MartinBasti: As I told, use property: 
```python
@property
def bundle():
    if self.nested:
        raise ...
   return format_bundle(self.bundle_dict)
```

operatorcourier/verify.py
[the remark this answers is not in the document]
kevinrizza: @jeremylinlin sure, but we should follow a particular style across the entire codebase. If we want to introduce a change like this, we should do it across every single function in the entire package -- and that is outside the scope of this PR.

operatorcourier/verify.py
kevinrizza: What? No. I'm saying that the purpose of the `validation_dict` was explicitly just to optionally print those logs to a file, right? So if we are mutating that object to do other things, that seems wrong. And if we aren't, we don't need to return it we can just do the file write in `build_and_verify()`

jeremylinlin: @kevinrizza But we do need to "mutate" the `validation_dict` which is initialized as an empty dict object with `warnings` and `errors` key. Previously, there is only 1 `validation_dict` because there's only 1 bundle. Now we have multiple bundles and therefore we have multiple those `validation_dict`s. The purpose of `get_validation_dict_from_manifests` method is to combine all those dicts into one single `validation_dict`, so that `build_and_verify` can write the whole thing into one file.

kevinrizza: We can still do this without explicitly returning it in the object. Just make the field effectively "private" and then create a method on the class that takes a path and writes the validation to a file. Then call that method inside `build_and_verify()`

operatorcourier/verify.py
jeremylinlin: Thanks @MartinBasti. Updated.

operatorcourier/verify.py
jeremylinlin: Thanks @kevinrizza . Updated.

kevinrizza:
@jeremylinlin is this ready to be reviewed again?

jeremylinlin:
@kevinrizza Yes!

jeremylinlin:
@MartinBasti Could you please also review and see if there's more changes needed? Thanks!

operatorcourier/verified_manifest.py
```diff
@@ -0,0 +1,153 @@
...
+                validation_info=validation_dict)
+
+        if not self.nested:
+            self.bundle_dict = bundle_dict
+
+        return validation_dict
+
+    def write_validation_to_file(self, file_path):
+        with open(file_path, 'w') as f:
+            f.write(json.dumps(self.__validation_dict) + "\n")
```
MartinBasti: Is better to python memory management to do:
```
f.write(json.dumps(...))
f.write('\n')
```

jeremylinlin: Thanks @MartinBasti. Updated.

PR #110, opened 2019-04-22
Catch correct error for invalid json in alm-examples

jeremylinlin:


operatorcourier/validate.py
```diff
@@ -289,7 +289,7 @@ def _csv_metadata_validation(self, metadata):
             if "alm-examples" in annotations:
                 try:
                     json.loads(annotations["alm-examples"])
-                except KeyError:
+                except json.decoder.JSONDecodeError:
```
kevinrizza: Since it is possible that there are other errors thrown here (I assume) shouldn't we also have a generic base case?

jeremylinlin: @kevinrizza Thanks for the comment. It makes sense. Should I just change `json.decoder.JSONDecodeError` to the generic `Exception`?

kevinrizza: I don't think there's anything inherently wrong with using the specific exception type if we are going to do something special there and then the generic `except:` when that does not happen. You can do:

```
try:
    ...
except json.decoder.JSONDecodeError:
    ...
except:
    ...
```

But that's only valuable if you are going to do something special with the result in that specific exception case.

jeremylinlin: @kevinrizza Yeah I get what you said and I think it makes sense. I was just confirming with you. I have made the update.

PR #111, opened 2019-04-22
nest: Fix error for operators without CRDs

SamiSousa:
- Add checks for object pathes in nest.py before accessing values
- Add test case for test_nest.py for operator without CRDs

Fixes an issue seen in CI on this PR: https://github.com/operator-framework/community-operators/pull/278
The CI runs `operator-courier nest` on operators, but for this operator it errors out with the obscure message`'customresourcedefinitions'`. This fix adds more field checks into the nest function, and handles operators without owned CRDs.
CI log: https://travis-ci.com/operator-framework/community-operators/builds/108167507#L454

### DIFF
--- a/operatorcourier/nest.py
+++ b/operatorcourier/nest.py
@@ -58,18 +58,23 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
             yaml.dump(csv, outfile, default_flow_style=False)
             outfile.flush()
 
-        csv_crds = csv["spec"]["customresourcedefinitions"]["owned"]
-        for csv_crd in csv_crds:
-            crd_name = csv_crd["name"]
-            if crd_name in crds:
-                crd = crds[crd_name]
-                crdfile_name = os.path.join(csv_folder, '%s.crd.yaml' % crd_name)
-                with open(crdfile_name, 'w') as outfile:
-                    yaml.dump(crd, outfile, default_flow_style=False)
-                    outfile.flush()
-            else:
-                errors.append("CRD %s mentioned in CSV %s was not found in directory."
-                              % (crd_name, csv_name))
+        if "customresourcedefinitions" in csv["spec"]:
+            if "owned" in csv["spec"]["customresourcedefinitions"]:
+                csv_crds = csv["spec"]["customresourcedefinitions"]["owned"]
+                for csv_crd in csv_crds:
+                    if "name" not in csv_crd:
+                        errors.append("CSV %s has an owned CRD without a `name` field"
+                                      "defined" % csv_name)
+                    crd_name = csv_crd["name"]
+                    if crd_name in crds:
+                        crd = crds[crd_name]
+                        crdfile_name = os.path.join(csv_folder, '%s.crd.yaml' % crd_name)
+                        with open(crdfile_name, 'w') as outfile:
+                            yaml.dump(crd, outfile, default_flow_style=False)
+                            outfile.flush()
+                    else:
+                        errors.append("CRD %s mentioned in CSV %s was not found in"
+                                      "directory." % (crd_name, csv_name))
 
     # if no errors were encountered, lets create the real directory and populate it.
     if len(errors) == 0:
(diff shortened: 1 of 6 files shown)


operatorcourier/nest.py
```diff
@@ -58,18 +71,23 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
...
-            if crd_name in crds:
-                crd = crds[crd_name]
-                crdfile_name = os.path.join(csv_folder, '%s.crd.yaml' % crd_name)
-                with open(crdfile_name, 'w') as outfile:
-                    yaml.dump(crd, outfile, default_flow_style=False)
-                    outfile.flush()
-            else:
-                errors.append("CRD %s mentioned in CSV %s was not found in directory."
-                              % (crd_name, csv_name))
+        if "customresourcedefinitions" in csv["spec"]:
```
jeremylinlin: `KeyError` here if `spec` is not defined?

operatorcourier/nest.py
```diff
@@ -46,7 +52,14 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
     # now lets create a subdirectory for each version of the csv,
     # and add all the relevant crds to it
     for csv in csvs:
+        if "metadata" not in csv:
+            errors.append("CSV has no `metadata` field defined")
         csv_name = csv["metadata"]["name"]
```
jeremylinlin: What about the `name` field?

operatorcourier/nest.py
```diff
@@ -46,7 +52,14 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
...
+
+        if "spec" not in csv:
+            errors.append("CSV %s has no `spec` field defined" % csv_name)
+        if "version" not in csv["spec"]:
+            errors.append("CSV %s has no `spec.version` field defined" % csv_name)
         version = csv["spec"]["version"]
```
jeremylinlin: `KeyError` here if `spec` or `version` is not defined?

operatorcourier/nest.py
```diff
@@ -58,18 +71,23 @@ def nest_bundles(yaml_files, registry_dir, temp_registry_dir):
...
-                errors.append("CRD %s mentioned in CSV %s was not found in directory."
-                              % (crd_name, csv_name))
+        if "customresourcedefinitions" in csv["spec"]:
+            if "owned" in csv["spec"]["customresourcedefinitions"]:
+                csv_crds = csv["spec"]["customresourcedefinitions"]["owned"]
+                for csv_crd in csv_crds:
+                    if "name" not in csv_crd:
+                        errors.append("CSV %s has an owned CRD without a `name` field"
+                                      "defined" % csv_name)
+                    crd_name = csv_crd["name"]
```
jeremylinlin: Same issue as above. I guess without using an `else` condition, the code will still try to access the `name` field when it is not present and will throw an error here?

operatorcourier/nest.py
SamiSousa: Good catch, will add a check here

operatorcourier/nest.py
SamiSousa: I append an error earlier on for this, but it's true that this would throw a `KeyError`
Should I break early if a key is missing, or should I continue to aggregate errors?

SamiSousa: I'm leaning towards using `continue` whenever a required key isn't found

SamiSousa:
@kevinrizza Addressed your comment, separated this PR into two commits relating to the distinct changes.
@jeremylinlin Addressed your KeyError concerns by adding `continue` statements in the CSV loop to avoid accessing keys that don't exist.
PTAL

jeremylinlin:
@SamiSousa Thanks for addressing the comments. It looks good!

tests/test_files/bundles/nest/bundle2/svcat.package.yaml
```diff
@@ -0,0 +1,5 @@
+packageName: svcat
+channels:
+- name: alpha
+  currentCSV: svcat.v0.1.34
+
```
kevinrizza: This is missing a newline at the end of the file.

SamiSousa: Ah, I left some whitespace on line 5, will fix this. Thank you!

SamiSousa:
@kevinrizza Addressed your comment, thank you!

PR #112, opened 2019-04-23
Add validations for customresourcedefinitions and apiservicedefinitions

darkowlzz:
Adds `_csv_crd_validation()` and `_csv_asd_validation` for validating `csv.spec.customresourcedefinitions` and `csv.spec.apiservicedefinitions` respectively.
Also validates the descriptors in both CRD and ASD using a common method `_csv_descriptors_validation()`.


operatorcourier/validate.py
```diff
@@ -184,79 +184,180 @@ def _csv_spec_validation(self, spec, bundleData):
...
+    def _csv_crd_validation(self, customresourcedefinitions, bundleData):
+        valid = True
+
+        crdList = []
+        for crd in bundleData[self.crdKey]:
+            try:
+                name = crd["metadata"]["name"]
+                crdList.append(name)
+            except KeyError:
+                pass
```
SamiSousa: Could this log a warning or error informing the user that a CRD is missing the `metadata.name` field? It's better to be noisy than quiet in this case (unless we previously warn elsewhere)

operatorcourier/validate.py
```diff
@@ -184,79 +184,180 @@ def _csv_spec_validation(self, spec, bundleData):
...
+
+        if "owned" not in customresourcedefinitions:
+            self._log_error("spec.customresourcedefinitions.owned"
```
SamiSousa: Not all operators that have the `customresourcedefinitions` field operate on `"owned"` CRDs. I believe this should be an early exit rather than a logged error

SamiSousa: I agree that having `customresourcedefinitions` without any subfields should raise an error. But there are other values than `"owned"` that could exist here (such as `required`: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Documentation/design/building-your-csv.md#required-crds)

operatorcourier/validate.py
```diff
@@ -184,79 +184,180 @@ def _csv_spec_validation(self, spec, bundleData):
...
+
+        return valid
 
-            if "owned" not in customresourcedefinitions:
-                self._log_error("spec.customresourcedefinitions.owned"
-                                "not defined for csv")
+    def _csv_asd_validation(self, apiservicedefinitions):
+        valid = True
+
+        if "owned" not in apiservicedefinitions:
```
SamiSousa: An operator may also have `"required"` apiservicedefinitions: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/Documentation/design/building-your-csv.md#required-apiservices

operatorcourier/validate.py
```diff
@@ -312,6 +413,31 @@ def _csv_spec_install_validation(self, install):
 
         return valid
 
+    def _csv_descriptors_validation(self, descriptors, crdName):
```
SamiSousa: Could we also check that the `x-descriptors` (if it exists) field is a list of strings?
Eq: https://github.com/operator-framework/community-operators/blob/master/community-operators/cluster-logging/cluster-logging.v4.1.0.clusterserviceversion.yaml#L275
Not sure if this is a requirement, more like an observation.

tests/test_files/yaml_source_dir/valid_yamls_with_single_crd/dynatrace-monitoring.v0.2.0.clusterserviceversion.yaml
```diff
@@ -38,7 +38,8 @@ metadata:
   name: dynatrace-monitoring.v0.2.0
   namespace: "placeholder"
 spec:
-  apiservicedefinitions: {}
+  apiservicedefinitions:
+    owned: {}
```
SamiSousa: Are we requiring that if the `apiservicedefinitions` or `customresourcedefinitions` fields exist that they have `owned` as a sub field (or the presence of any valid sub field)?
(Also this should be a list not an object)

SamiSousa: Is our stance that if it exists and is empty that is valid? If so, these both should be valid:
```
spec:
  apiservicedefinitions: {}
```
and
```
spec:
  apiservicedefinitions:
    owned: []
```

SamiSousa: I want to clarify that my stance is based on my observations of the existing community-operators CSVs, many of which set `spec.apiservicedefinitions` to the empty object `{}`.
We *could* enforce that the key `owned` exists, but should we/does it create any benefit?

operatorcourier/validate.py
SamiSousa: > I agree that having `customresourcedefinitions` without any subfields should raise an error.

To contradict myself, what's the benefit of enforcing that subfields exist?

tests/test_files/yaml_source_dir/valid_yamls_with_single_crd/dynatrace-monitoring.v0.2.0.clusterserviceversion.yaml
darkowlzz: Thanks. This was a mistake based on requiring `owned`. `owned` is not required, and this change is not needed.

flickerfly:
/test 

Because I'm interested in this work and it has been awhile.

PR #113, opened 2019-04-24
Raise error after validation is written to file

jeremylinlin:
Currently when we initialize the `VerifiedManifest` object and validate the input yamls, if the validation fails, `OpCourierBadBundle` Error will be raised before the warnings/errors are written to a file (if the user specifies this option).

We need to move the code raising the error after the warnings/errors have been written to a file.

### DIFF
--- a/operatorcourier/api.py
+++ b/operatorcourier/api.py
@@ -14,6 +14,7 @@ from operatorcourier.verified_manifest import VerifiedManifest
 from operatorcourier.push import PushCmd
 from operatorcourier.nest import nest_bundles
 from operatorcourier.flatten import flatten_bundles
+from operatorcourier.errors import OpCourierBadBundle
 
 logger = logging.getLogger(__name__)
 
@@ -49,6 +50,10 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
 
     if validation_output:
         verified_manifest.write_validation_to_file(validation_output)
+
+    if not verified_manifest.is_valid:
+        raise OpCourierBadBundle("Resulting bundle is invalid, "
+                                 "input yaml is improperly defined.", {})
 
     return verified_manifest
 
--- a/operatorcourier/verified_manifest.py
+++ b/operatorcourier/verified_manifest.py
@@ -30,6 +30,7 @@ class VerifiedManifest:
         self.bundle_dict = None
         self.__validation_dict = \
             self.get_validation_dict_from_manifests(manifests, ui_validate_io, repository)
+        self.is_valid = False if self.__validation_dict['errors'] else True
 
     def get_manifest_files_content(self, file_paths):
         """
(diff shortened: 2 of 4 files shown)


jeremylinlin:
@SamiSousa It seems [those](https://github.com/operator-framework/operator-courier/blob/5d3fafa3ac8da0109904880e4df120c38519cd92/operatorcourier/verified_manifest.py#L104-L110) are the only 2 places where the "other potential errors" are. But we probably should not also put them into the `validation_dict`, since that dict comes from `ValidateCmd` class that validate only the yaml content. For those other errors, I guess we probably should just raise the error and exit, just like [here](https://github.com/operator-framework/operator-courier/blob/5d3fafa3ac/operatorcourier/api.py#L46).

operatorcourier/api.py
```diff
@@ -50,6 +50,8 @@ def build_and_verify(source_dir=None, yamls=None, ui_validate_io=False,
     if validation_output:
         verified_manifest.write_validation_to_file(validation_output)
 
+    verified_manifest.raise_error_if_input_invalid()
```
kevinrizza: It seems like imperative error handling is the wrong way to do this. Why can't we just raise the error when we encounter it during the validation?

jeremylinlin: @kevinrizza You mean the errors encountered in `validate.py`? I'm just following the way we do [previously](https://github.com/operator-framework/operator-courier/blob/ed89aded2c/operatorcourier/api.py#L73). For the validation errors, we collect all of them during validation, write them to a file if `validation_output` flag is specified, and raise the error if the validation fails.

But now the error is raised before we are able to write it to a file. I'm just changing the order to align how we do it previously.

kevinrizza: @jeremylinlin Before we were checking a property and then doing the failure ourselves in the API method. Now it is on the implementation of a different method (which does not have any sort of contract or return value) to raise an exception. It seems like either the process of calling validate should throw the error, or at the very least return some value that says whether or not there is an error. We did the latter before, and we are not doing it anymore with this change.

jeremylinlin: @kevinrizza The problem is that we do not want to expose `self.__validation_dict` in the API method. Previously we only return the `bundle`, so having `validate()` return `validation_results_dict` does not have an issue. If we now have the `verified_manifest` object expose the dict to the API method, then users will have the access to it from the `verified_manifest` object, making the private field `self.__validation_dict` meaningless.

kevinrizza: @jeremylinlin we don't need to use the dictionary to do the validation -- we just need a boolean to tell us if the bundle is valid or not. We used to also return a bool `valid` when we called validate. Why can't we continue to do something like that?

kevinrizza:
@jeremylinlin Can this also include a test so that we don't regress here again?

operatorcourier/api.py
jeremylinlin: @kevinrizza Oh because we also need the [dict](https://github.com/jeremylinlin/operator-courier/blob/raise-error-after-output-validation-yaml/operatorcourier/verified_manifest.py#L155) to raise the error.

kevinrizza: @jeremylinlin I'm not saying don't have a private method that figures out what to do when we need to know to write the data for the error. I'm saying that the actual check to see if an error needs to be thrown, and the throw itself, should be in the API method. Like:

```
if not valid:
    ValidationManifest.__write_validation_error()
    raise Exception
```

Otherwise you are relying on a private field in a class function to determine the control flow of the API method itself. Which, if that method was changed, could result in unexpected behavior in a future refactor/enhancement. If the error itself is checked for and thrown in the API method, then at least the intent is much clearer.

jeremylinlin:
@kevinrizza Thanks for the comments. I have addressed them and also added several integration tests so that similar bugs can be captured in the future.

PR #114, opened 2019-04-25
Update version to 2.0.0

kevinrizza:


jeremylinlin:
@kevinrizza Maybe we should wait until this [PR](https://github.com/operator-framework/operator-courier/pull/113) is merged? Right now the `validation_output` flag does not work when there's an error.

kevinrizza:
> @kevinrizza Maybe we should wait until this [PR](https://github.com/operator-framework/operator-courier/pull/113) is merged? Right now the `validation_output` flag does not work when there's an error.

Agreed, just preparing the PR so all are aware that we want to release.

PR #118, opened 2019-04-29
Remove SemVer requirement for nested version folders for flatten command

jeremylinlin:
This PR solves similar problem of https://github.com/operator-framework/operator-courier/pull/116, where we do not strictly require the version folder name to be a SemVer.

However, we still need the SemVer of the directory in order to compare and retrieve the latest version of each CRD from each version folder. We get this SemVer by first parsing the `spec.version` field in CSV of each folder.


operatorcourier/flatten.py
```diff
@@ -61,34 +63,56 @@ def get_flattened_files_info(source_dir: str) -> [(str, str)]:
...
-    try:
-        semver.parse(version_folder_name)
-    except ValueError:
-        logger.warning("Ignoring %s as it is not a valid semver. "
-                       "See https://semver.org for the semver specification.",
-                       version_folder_name)
-        return
-
-    logger.info('Parsing folder: %s...', version_folder_name)
+    logger.info('Parsing folder %s with semver %s', version_folder_name, folder_semver)
```
kevinrizza: Can this be `Parsing folder %s for operator version %s` ?

operatorcourier/flatten.py
```diff
@@ -61,34 +63,56 @@ def get_flattened_files_info(source_dir: str) -> [(str, str)]:
...
+            except KeyError:
+                msg = f'{item} is not a valid CSV file as "spec.version" ' \
+                      f'field is required'
+                logger.error(msg)
+                raise errors.OpCourierBadBundle(msg, {})
+            return csv_version
+
+    msg = 'This version directory does not contain any valid CSV file.'
+    logger.error(msg)
+    raise errors.OpCourierBadBundle(msg, {})
```
kevinrizza: If we're already going through this trouble, can't we just return nil here? Why do we need to throw an error if we have some extra folders we don't care about?

ralphbean:
Any progress on this one @jeremylinlin - @shawn-hurley's build is blocked on this and we'd like to be able to get it to QE.

jeremylinlin:
Hi @ralphbean, sorry for the delay. I have just updated the PR @kevinrizza.

operatorcourier/flatten.py
```diff
@@ -61,34 +65,55 @@ def get_flattened_files_info(source_dir: str) -> [(str, str)]:
...
+                logger.error(msg)
+                raise errors.OpCourierBadYaml(msg)
+            except KeyError:
+                msg = f'{item} is not a valid CSV file as "spec.version" ' \
+                      f'field is required'
+                logger.error(msg)
+                raise errors.OpCourierBadBundle(msg, {})
+            return csv_version
+
+    return ''
```
MartinBasti: Is better to return `None` for nothing

kevinrizza:
lgtm aside from Martin's one comment

csomh:
:heavy_check_mark: flattening and pushing the manifests from the blocked build works with this patch.

operatorcourier/flatten.py
ralphbean: @jeremylinlin, can you address this?

jeremylinlin: @ralphbean Addressed

PR #121, opened 2019-05-01
Run in docker

reuvenharrison:
Amended readme with option to run in docker.
It's useful when you want to automate the artifact validation in a CI pipeline and you don't have Python 3.6 on the VM/container that is running the pipeline.
The docker wrapper is here: https://github.com/Tufin/operator-courier
The docker image is here: https://hub.docker.com/r/tufin/operator-courier


kevinrizza:
@reuvenharrison This is nice. Could you add the dockerfile to the PR so that it could be built and amended as part of the release process?

kevinrizza:
Also please squash commits.

reuvenharrison:
Applied your feedback.
Please remember to update the image name in the readme after you build it yourself (currently points to the image I built).

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM python:3
+
+RUN pip3 install operator-courier
```
MartinBasti: Could operator-courier be installed from the local dir `pip3 install .`  It allows to build containers directly from git source automatically.

kevinrizza:
@reuvenharrison Sorry, I was unclear in my previous request. I meant that, instead of pointing at all to your locally built image directly, could you just provide the dockerfile and some documentation around building and using it locally? Then, if it works well we can create a followup pr with the image built as part of the build process.

I also liked @MartinBasti 's suggestion about pulling down the source and creating the binary at build time, since then we can version the container image rather than always having the latest at runtime.

reuvenharrison:
Changed Dockerfile to install operator-courier from the local dir.
To build the image: 
```docker build operator-courier .```
To use the container:
```docker run <image-name> operator-courier ...```

File: README.md [hunk 2]
```diff
@@ -95,4 +95,9 @@ $ tox
 This will run the tests with several versions of Python 3, measure coverage,
 and run flake8 for code linting.
 
+## Run in Docker
+```
+docker run tufin/operator-courier operator-courier
```
kevinrizza: I would prefer, based on your latest change, that the README did not point to a particular docker image. Instead it should describe how to build the docker image, i.e.:

```
docker build Dockerfile -t $TAG
docker run $TAG operator-courier
```

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:3
+
+ADD . /repo
+WORKDIR /repo
+RUN pip3 install .
+RUN rm -rf /repo
+
+CMD [ "operator-courier", "--help"]
```
kevinrizza: Please add a newline at the end of this file.

reuvenharrison: done

Dockerfile
reuvenharrison: done

README.md
reuvenharrison: done

PR #122, opened 2019-05-01
Only parse manifest folders with valid bundle files

jeremylinlin:
We previously require the nested manifest directories to have a SemVer name. But this blocks builds from another team that uses courier in their pipeline.

So we need to relax that requirement and parse all directories even if it does not have a SemVer name.

The requirement is removed in https://github.com/operator-framework/operator-courier/pull/116 and https://github.com/operator-framework/operator-courier/pull/118 when we verify manifest directory and run `flatten` command. But we need to further refactor it to only parse valid manifest folders, and ignore the rest (instead of raising errors when invalid folders are parsed). In the meantime, we also need to properly identify whether the input source directory structure is flat or nested after parsing all directories and files, and perform different actions accordingly.

## APIs affected:
- `build_and_verify(source_dir, ...)`
- `build_verify_and_push(source_dir, ...)`
- `flatten`

## CLI subcommands affected
- `verify`
- `push`
- `flatten`

## Valid Nested Layout
- Previously, all nested manifest directories will be parsed and the API does not check if they contain valid manifest files until invalid files are encountered and errors are raised.
- Now, only the manifest directories with valid manifest files (meaning at least 1 valid CSV file is present inside the directory) will be parsed, and only valid CSV and CRD files will be extracted from the directories.
- If the source directory satisfies both nested and flat structure, meaning both valid manifest directories and valid CSV+package file are found in source directory root, we parse the source directory as it is in nested layout, and ignores the CSV in source root.

## Valid Flat Layout
- Previously, as long as there are subdirectories found in source_dir, we think of the structure as nested.
- Now, since all subdirectories are processed first. If no **valid** manifest directories are found in source_dir, but there are still valid CSV+package file found in source_dir, then we will consider it to be a valid flat structure.


kevinrizza:
@jeremylinlin Your PR description describes the limitations of what is happening now in terms of how you wanted the code to be refactored, but it does not actually describe *what* has changed from the perspective of how the API is affected. Could you please explain in your description (and commit message) which edge cases this PR is attempting to resolve, and what is expected from this?

jeremylinlin:
@kevinrizza Thanks for the comment! I've updated the description and commit message.

operatorcourier/flatten.py
```diff
@@ -4,8 +4,10 @@
 from typing import Dict, Tuple
 from shutil import copyfile
 import semver
-from operatorcourier import identify
-from operatorcourier import errors
+from operatorcourier.identify import get_operator_artifact_type
+from operatorcourier.errors import OpCourierBadBundle, OpCourierBadYaml
+from operatorcourier.verified_manifest \
+    import is_manifest_folder, get_package_csv_info_from_root
```
kevinrizza: This seems like a little bit of an anti pattern. If `Verified_Manifest` is a class, shouldn't only class functions come from this package? If they aren't defined as part of this class (which would imply they were called by an instance of that type) then they probably shouldn't be in that module

operatorcourier/flatten.py
```diff
@@ -22,99 +24,101 @@ def get_flattened_files_info(source_dir: str) -> [(str, str)]:
...
     """
-
-    # extract package file and version folders from source_dir
-    root, folder_names, file_names = next(os.walk(source_dir))
-    if not folder_names:
+    root_path, dir_names, root_dir_files = next(os.walk(source_dir))
+    package_content, contains_csv = get_package_csv_info_from_root(source_dir)
+
+    dir_paths = [os.path.join(source_dir, dir_name) for dir_name in dir_names]
+    manifest_paths = list(filter(lambda x: is_manifest_folder(x), dir_paths))
```
kevinrizza: Can we add a comment here about this code? We had ` extract package file and version folders from source_dir`, and it was removed without any comment explaining what this code does now.

operatorcourier/identify.py
```diff
@@ -20,17 +20,14 @@ def get_operator_artifact_type(operatorArtifactString):
         logger.error(msg)
         raise OpCourierBadYaml(msg)
     else:
-        artifact_type, artifact_name = None, None
+        artifact_type = None
```
kevinrizza: This change seems unrelated to the purpose of the pr. I agree that it is a valid change, but it should be separated into a separate pr.

operatorcourier/verified_manifest.py
```diff
@@ -9,6 +9,81 @@
 
 logger = logging.getLogger(__name__)
 FLAT_KEY = '__flat__'
+CRD_KEY = 'CustomResourceDefinition'
+CSV_KEY = 'ClusterServiceVersion'
+
+
+def is_manifest_folder(folder_path):
```
kevinrizza: See my comment above about how it seems like these helper functions don't belong in the same module as the verified_manifest class module.

operatorcourier/verified_manifest.py
```diff
@@ -20,81 +95,70 @@ def bundle(self):
         return format_bundle(self.bundle_dict)
 
     def __init__(self, source_dir, yamls, ui_validate_io, repository):
+        self.nested = False
```
kevinrizza: Is this needed? Shouldn't this boolean be defined as `False` by default?

operatorcourier/verified_manifest.py
```diff
@@ -20,81 +95,70 @@ def bundle(self):
...
-        Given a source directory OR a list of yaml files. The function returns a dict
-        containing all operator bundle file information grouped by version. Note that only
-        one of source_dir or yamls can be specified.
-
-        :param source_dir: Path to local directory of operator bundles, which can be
-                           either flat or nested
-        :param yamls: A list of yaml strings to create bundle with
-        :return: A dictionary object where the key is the semantic version of each bundle,
-                 and the value is a list of yaml strings of operator bundle files.
+        Given a source directory, the method returns a dict containing all
```
kevinrizza: nit: `this method` instead of `the method`

operatorcourier/identify.py
jeremylinlin: Thanks @kevinrizza. Removed the change and created a PR https://github.com/operator-framework/operator-courier/pull/123 for this change.

operatorcourier/verified_manifest.py
jeremylinlin: @kevinrizza I think it is needed because unlike Java, Python does not have a way to explicitly declare a variable with a type that has a default value.

If I remove this line, my IDE will give a warning at this [line](https://github.com/operator-framework/operator-courier/pull/122/files#diff-9bab59f69ef6c22e41f4c0d543b8e228R138) where the boolean is used, saying **instance attribute defined outside \_\_init\_\_**. I guess that's a bad practice in Python?

jeremylinlin:
@kevinrizza Thanks for the comments! I have addressed them and please take a look for new changes.

operatorcourier/verified_manifest.py
kevinrizza: Good point.

operatorcourier/parse_manifests.py
```diff
@@ -0,0 +1,82 @@
+import os
```
kevinrizza: Can we name this file a noun instead of a verb? Like `manifest_helper` or even `manifest_parser` ?

jeremylinlin: @kevinrizza Done.

jeremylinlin:
Hey @MartinBasti, could you please also review it when you get a chance? Thank you!

operatorcourier/flatten.py
```diff
@@ -22,99 +24,104 @@ def get_flattened_files_info(source_dir: str) -> [(str, str)]:
...
+    # get package content and check if CSV exists in source_dir root, and
+    # process all subdirectories and filter those that do not contain valid manifest files
+    root_path, dir_names, root_dir_files = next(os.walk(source_dir))
+    package_content, contains_csv = get_package_csv_info_from_root(source_dir)
+
+    dir_paths = [os.path.join(source_dir, dir_name) for dir_name in dir_names]
+    manifest_paths = list(filter(lambda x: is_manifest_folder(x), dir_paths))
+
+    # nested layout
+    if len(manifest_paths) > 0:
```
MartinBasti: `if manifests_paths:` should be enough

operatorcourier/manifest_parser.py
```diff
@@ -0,0 +1,82 @@
+import os
+import logging
+from operatorcourier.errors import OpCourierBadBundle
+from operatorcourier import identify
```
MartinBasti: in `flatten.py` you refactored this import:

```diff
-from operatorcourier import identify
+from operatorcourier.identify import get_operator_artifact_type
```

and made review harder by this non-relevant change. And suddenly here you use inconsistently the same import you refactored in previous files. So when you refactor keep it consistent.

operatorcourier/manifest_parser.py
```diff
@@ -0,0 +1,82 @@
...
+    :return: a dict where the key is either  'ClusterServiceVersion' or
+             'CustomResourceDefinition', and the value is a list of the valid
+             CSV and CRD files content
+    """
+    crd_csv_files_content = {}
+    for item in os.listdir(folder_path):
+        item_path = os.path.join(folder_path, item)
+        if not os.path.isfile(item_path):
+            continue
+        if item.endswith(".yaml") or item.endswith(".yml"):
```
MartinBasti: This code repeats everywhere, would be nice to factor it out into separate function `is_yaml_file`

operatorcourier/verified_manifest.py
```diff
@@ -20,81 +22,70 @@ def bundle(self):
...
+        package_content, contains_csv = get_package_csv_info_from_root(source_dir)
+
+        dir_paths = [os.path.join(source_dir, dir_name) for dir_name in dir_names]
+        manifest_paths = list(filter(lambda x: is_manifest_folder(x), dir_paths))
+
+        # if there is at least 1 valid manifest folder, we treat the directory layout as
+        # nested, and add all manifest files from each version folder to manifests dict
+
+        # nested layout: add package to each manifest dict entry
+        if len(manifest_paths) > 0:
```
MartinBasti: `if manifests_paths:`

operatorcourier/manifest_parser.py
```diff
@@ -0,0 +1,82 @@
...
+            with open(item_path) as f:
+                file_content = f.read()
+            if CSV_KEY == identify.get_operator_artifact_type(file_content):
+                return True
+
+    folder_name = os.path.basename(folder_path)
+    logging.warning('Ignoring folder "%s" as it is not a valid manifest '
```
MartinBasti: Did you mean `logger.warning` ?

tests/integration/test_push.py
```diff
@@ -46,12 +46,12 @@ def ensure_application_release_removed(repository_name, release_version):
     ('tests/test_files/yaml_source_dir/invalid_yamls_without_package',
      'oneagent',
      '0.0.1',
-     'ERROR:operatorcourier.validate:Bundle does not contain any packages.'),
+     'ERROR:root:Bundle does not contain any packages.'),
```
MartinBasti: Why `root` logger is used?

tests/integration/test_verify.py
```diff
@@ -87,10 +92,10 @@ def test_verify_invalid_nested_sources_with_output(source_dir):
 
 @pytest.mark.parametrize('source_dir,error_message', [
     ("tests/test_files/yaml_source_dir/invalid_yamls_without_package",
-     "ERROR:operatorcourier.validate:Bundle does not contain any packages."),
+     "ERROR:root:Bundle does not contain any packages."),
```
MartinBasti: why root logger is used?

operatorcourier/manifest_parser.py
```diff
@@ -0,0 +1,82 @@
...
+        if root_file_path.endswith(".yaml") or root_file_path.endswith(".yml"):
+            with open(root_file_path) as f:
+                file_content = f.read()
+            file_type = identify.get_operator_artifact_type(file_content)
+            if file_type == 'ClusterServiceVersion':
+                contains_csv = True
+            elif file_type == 'Package':
+                if package_content:
+                    msg = 'Only 1 package is expected to exist in source root folder.'
+                    logging.error(msg)
```
MartinBasti: here

operatorcourier/manifest_parser.py
```diff
@@ -0,0 +1,82 @@
...
+                    raise OpCourierBadBundle(msg, {})
+                package_content = file_content
+
+    if not package_content:
+        msg = 'Bundle does not contain any packages.'
+        logging.error(msg)
```
MartinBasti: and here

operatorcourier/verified_manifest.py
```diff
@@ -20,81 +22,70 @@ def bundle(self):
...
+            self.nested = True
+            logging.info('The source directory is in nested structure.')
```
MartinBasti: and here

operatorcourier/verified_manifest.py
```diff
@@ -20,81 +22,70 @@ def bundle(self):
...
-                    package_content = file_content
-            if not package_content:
-                msg = 'No package file exists in the nested bundle.'
-                logging.error(msg)
-                raise OpCourierBadBundle(msg, {})
-            for version in manifests:
-                manifests[version].append(package_content)
+            msg = 'The source directory structure is not in valid flat or nested format,'\
+                  'because no valid CSV file is found in root or manifest directories.'
+            logging.error(msg)
```
MartinBasti: and here

operatorcourier/manifest_parser.py
jeremylinlin: @MartinBasti Changed `from operatorcourier.identify import get_operator_artifact_type` back to `from operatorcourier import identify`.

jeremylinlin:
@MartinBasti Thank you very much for the comments! I have addressed all of them and please take a look when you are free.

PR #123, opened 2019-05-02
Remove logging for identify

jeremylinlin:
Because we call this `get_operator_artifact_type()` method multiple times at different places, this will create a lot of noise to log messages which are not necessary. Removing it makes it cleaner to the users.


jeremylinlin:
@SamiSousa Thanks for reviewing my PR!

PR #125, opened 2019-05-06
Refactor nest command to support both flat and nested structures

jeremylinlin:
The purpose of this PR is similar to PR https://github.com/operator-framework/operator-courier/pull/122.

Previously, we assume the input source directory for `nest` command is a flat directory. Therefore it fails when the input directory is already nested.

Now we want the `nest` command to work in both **flat** and **nested** structures. And because now we do not strictly require the input source directory to only contain manifest file and folders, we also need to make sure the extra files and folders are filtered when we copy files to the output directory.


operatorcourier/api.py
```diff
@@ -99,28 +99,26 @@ def build_verify_and_push(namespace, repository, revision, token,
...
 
-def nest(source_dir, registry_dir):
+def nest(source_dir, output_dir):
     """Nest takes a flat bundle directory and version nests it
     to eventually be consumed as part of an operator-registry image build.
 
+    This method will only extract valid operator manifest files and folders
+    and ignore the rest.
+
+    If the input directory is already nested, this method will copy the file and
```
kevinrizza: Nit: `copy the files` instead of `copy the file`

operatorcourier/api.py
```diff
@@ -99,28 +99,26 @@ def build_verify_and_push(namespace, repository, revision, token,
...
-
-    if source_dir is not None:
-        for filename in os.listdir(source_dir):
-            if filename.endswith(".yaml") or filename.endswith(".yml"):
-                with open(os.path.join(source_dir, filename)) as f:
-                    yaml_files.append(f.read())
-
-    with TemporaryDirectory() as temp_dir:
-        nest_bundles(yaml_files, registry_dir, temp_dir)
+    if source_dir:
```
kevinrizza: I'm confused by the purpose of this if statement. Before we used `if source_dir is not None:` because there was code directly below it that used the `source_dir` in os calls and it could behave badly if it wasn't set. Now though, we are just passing these parameters along to the `nest_bundles` function, so is this if statement necessary at all? Or, if we feel that it is necessary, shouldn't both parameters be checked since we are just ensuring that we got input?

operatorcourier/manifest_parser.py
```diff
@@ -31,16 +33,17 @@ def is_manifest_folder(folder_path):
...
-    Given a folder path, the method returns a dict containing manifest files
-    content grouped by CRD and CSV key.
+    Given a folder path, the method returns the CRD and CSV files info parsed from the
+    input directory.
     :param folder_path: the path of the input folder
-    :return: a dict where the key is either  'ClusterServiceVersion' or
-             'CustomResourceDefinition', and the value is a list of the valid
-             CSV and CRD files content
+    :return: CRD and CSV files info parsed from the input directory. Each files_info
+             is a list of tuple, where each tuple contains two elements, namely
```
kevinrizza: nit: `a list of tuples`

operatorcourier/manifest_parser.py
```diff
@@ -49,37 +52,54 @@ def get_crd_csv_files_content(folder_path):
...
 
-def get_package_csv_info_from_root(source_dir):
+
+def get_csvs_pkg_info_from_root(source_dir: str) -> Tuple[List[Tuple], Tuple]:
+    """
+    Given a source directory path, the method returns the CSVs and package file info
+    parsed from the input directory.
+    :param source_dir: the path of the input source folder
+    :return: CSVs and package file info parsed from the input directory.
+             csvs_info is a list of tuple whereas pkg_info is a single tuple, and
```
kevinrizza: nit: `a list of tuples`

jeremylinlin:
@kevinrizza Thanks for reviewing the PR, I have addressed your comments.
@MartinBasti Could you please review this PR when you get a chance? Thank you!

SamiSousa:
@kevinrizza @MartinBasti Are any additional changes needed for this PR to be merged?

operatorcourier/nest.py
```diff
@@ -2,34 +2,88 @@
...
+        for manifest_path in manifest_paths:
+            crds_info, csvs_info = get_crd_csv_files_info(manifest_path)
+            crd_csv_file_paths = [file_info[0] for file_info in (crds_info + csvs_info)]
+            manifest_files_path.extend(crd_csv_file_paths)
+
+        # copy all manifest files to output_dir with folder structure preserved
+        source_dir = os.path.join(source_dir, '')  # adds a trailing slash if missing
+        os.makedirs(output_dir, exist_ok=True)
+        for file_path in manifest_files_path:
+            file_path_relative = file_path[len(source_dir):]
```
MartinBasti: Could you try this? https://docs.python.org/3/library/os.path.html#os.path.relpath

You don't need this hack with that `source_dir = os.path.join(source_dir, '')  # adds a trailing slash if missing`

PR #133, opened 2019-05-28
Add filename to warnings and errors

kevinrizza:
Problem:
Command line output logs are hard to read because, in the case where
there are many operator yamls, there is no obvious output as to which
file caused the error in question.

Solution:
Add file references and relative file paths to the output logs.
Update build_bundle() to take the filename metadata so that the
validator can use the new metadata field

### DIFF
--- a/operatorcourier/verified_manifest.py
+++ b/operatorcourier/verified_manifest.py
@@ -25,7 +25,8 @@ class VerifiedManifest:
         self.nested = False
 
         if yamls:
-            manifests = {FLAT_KEY: yamls}
+            yaml_strings_with_metadata = self._set_empty_filepaths(yamls)
+            manifests = {FLAT_KEY: yaml_strings_with_metadata}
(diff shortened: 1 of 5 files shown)


operatorcourier/verified_manifest.py
```diff
@@ -34,6 +35,14 @@ def __init__(self, source_dir, yamls, ui_validate_io, repository):
             self.get_validation_dict_from_manifests(manifests, ui_validate_io, repository)
         self.is_valid = False if self.__validation_dict['errors'] else True
 
+    def _set_empty_filepaths(self, yamls):
```
awgreene: Could you explain what this function does?

kevinrizza: When you call the api with `yamls` set instead of a file path there are no file names. So this function just creates the tuples for the file list and sets them all to empty string.

PR #137, opened 2019-06-19
Ignore extra files in manifest dir

kevinrizza:
* Update the identifier to return empty string when parsing unexpected
file (i.e. non csv/crd/package)
* Update callers to ignore file when filetype returned is none
* Update tests to reflect this change

### DIFF
--- a/operatorcourier/build.py
+++ b/operatorcourier/build.py
@@ -16,10 +16,6 @@ class BuildCmd():
             )
         )
 
-    def _get_field_entry(self, yamlContent):
-        yaml_type = identify.get_operator_artifact_type(yamlContent)
-        return yaml_type[0:1].lower() + yaml_type[1:] + 's'
-
     def _get_relative_path(self, path):
         """
         :param path: the path of the file
@@ -31,13 +27,20 @@ class BuildCmd():
 
     def _updateBundle(self, operatorBundle, file_name, yaml_string):
         # Determine which operator file type the yaml is
-        operator_artifact = self._get_field_entry(yaml_string)
+        operator_artifact = identify.get_operator_artifact_type(yaml_string)
+
+        # If the file isn't one of our special types, we ignore it and return
+        if operator_artifact == identify.UNKNOWN_FILE:
+            return operatorBundle
+
+        # Get the array name expected by the dictionary for the given file type
+        op_artifact_plural = operator_artifact[0:1].lower() + operator_artifact[1:] + 's'
 
         # Marshal the yaml into a dictionary
         yaml_data = yaml.safe_load(yaml_string)
 
         # Add the data dictionary to the correct list
-        operatorBundle["data"][operator_artifact].append(yaml_data)
+        operatorBundle["data"][op_artifact_plural].append(yaml_data)
 
         # Encode the dictionary into a string, then use that as a key to reference
         # the file name associated with that yaml file. Then add it to the metadata.
--- a/operatorcourier/identify.py
+++ b/operatorcourier/identify.py
@@ -1,19 +1,24 @@
 from yaml import safe_load
 from yaml import MarkedYAMLError
 import logging
-from operatorcourier.errors import OpCourierBadYaml, OpCourierBadArtifact
+from operatorcourier.errors import OpCourierBadYaml
 from operatorcourier.manifest_parser import CRD_STR, CSV_STR, PKG_STR
 
 logger = logging.getLogger(__name__)
 
+UNKNOWN_FILE = "Unknown"
+
 
 def get_operator_artifact_type(operatorArtifactString):
     """get_operator_artifact_type takes a yaml string and determines if it is
-    one of the expected bundle types: ClusterServiceVersion,
-    CustomResourceDefinition, or Package.
+    one of the expected bundle types.
 
     :param operatorArtifactString: Yaml string to type check
     """
+
+    # Default to unknown file unless identified
+    artifact_type = UNKNOWN_FILE
+
     try:
         operatorArtifact = safe_load(operatorArtifactString)
     except MarkedYAMLError:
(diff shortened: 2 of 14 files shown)


operatorcourier/build.py
```diff
@@ -33,6 +35,10 @@ def _updateBundle(self, operatorBundle, file_name, yaml_string):
         # Determine which operator file type the yaml is
         operator_artifact = self._get_field_entry(yaml_string)
 
+        # If the file isn't one of our special types, we ignore it and return
+        if operator_artifact == "":
```
aravindhp: Alternate would be to allow including all resource types that OLM supports. Please see: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator.go#L1024

MartinBasti:
@kevinrizza tests fail

operatorcourier/identify.py
```diff
@@ -29,7 +29,5 @@ def get_operator_artifact_type(operatorArtifactString):
                 artifact_type = operatorArtifact["kind"]
             if artifact_type is not None:
                 return artifact_type
-
-        msg = 'Courier requires valid CSV, CRD, and Package files'
-        logger.error(msg)
-        raise OpCourierBadArtifact(msg)
+            else:
+                return ""
```
MartinBasti: What happen when `operatorArtifact` is not a dict?

Now it will return `None`. Should it raise an error instead?

Can we instead of empty string use `None` as representation that there is no operator_artifact?

operatorcourier/build.py
```diff
@@ -33,6 +35,10 @@ def _updateBundle(self, operatorBundle, file_name, yaml_string):
         # Determine which operator file type the yaml is
         operator_artifact = self._get_field_entry(yaml_string)
```
MartinBasti: Could we just do this here instead of rewriting all the code?
```python
try:
    yaml_type = identify.get_operator_artifact_type(yamlContent)
except OpCourierBadArtifact:
   # no valid artifact skipping update
   return operratorBundle

```

kevinrizza: My concern is that a future update is going to force us to support all files and actually bundle them. In that case, it seems like this method shouldn't return an error.

operatorcourier/build.py
kevinrizza: This list is going to keep growing, I assume? Is there a reason we shouldn't just accept any yaml file?

operatorcourier/identify.py
kevinrizza: +1, good point

kevinrizza:
> Great work Kevin, although I do have one concern - it seems like only the CRDs, CSVs, and Packages will be placed in the zip that is pushed to Quay. Should the yamls that [OLM supports](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/pkg/controller/operators/catalog/operator.go#L1024) be pushed as well?

Seems like it would be extremely difficult to keep these in sync. In that case, it might just be easier for us to bundle any yaml file that we find in the directory.

operatorcourier/build.py
aravindhp: OLM will fail the install if it finds an unsupported resource

operatorcourier/build.py
MartinBasti: I don't see difference in code flow. If `get_operator_artifact_type` doesn't recognize type, instead of `return` just bundle file anyway with a generic method.

kevinrizza: I think doing something like that is really twisting the purpose of a lot of these abstractions. The problem is that `build` used to just mean `create a dictionary that includes arrays of these files`, Now we are still doing that and then trying to apply a directory structure on top of it by unraveling the dictionary.

All this pr was intended to do is ignore extra files by not throwing an error. I do not see the value of continuing to throw an error here if we want to always be able to call this method on any arbitrary file. If we don't want to be concerned with "Is this `None`, `EmptyString`, or a real result" then it seems like the correct thing to do is create an enumeration with all the possible file types and include an item like "None of the Above" that returns when it isn't one of our expected files. And anywhere that calls that method needs to deal with that case as well.

operatorcourier/build.py
kevinrizza: It seems to me like that problem is slightly orthogonal to the issue we are trying to solve with this PR. We should address it in a followup.

aravindhp: I am fine with that. Just wanted to call this out in case it was low hanging fruit.

operatorcourier/build.py
MartinBasti: When function cannot parse artifact_type raising exception is IMHO the best thing what can it do in python. You got clear message that function failed and you can catch that exception and modify behavior of upper levels of code accordingly.

Why a some sub-sequential function cannot fail? Build method can catch exception and does different operation, so build method itself will never fail.

I'm fine with returning `None` if artifact_type cannot be parsed but it will adds there unnecessary `if artifact_type is None` lines

kevinrizza: I agree that the difference between returning none or empty string is problematic, so instead I have updated the method to have the following behavior:

Return the manifest file type if it can be determined
Return special Unknown type if it can't be determined
Throw an error if the yaml is invalid

In that way, all callers are forced to handle any behavior based on what the file type is (specific or unexpected) and have specific knowledge that the yaml is valid.

kevinrizza:
@awgreene @MartinBasti Please take another look, I updated the identify method to include a new result `identify.UNKNOWN_FILE` that will be the result for any valid yaml file that isn't specifically one of the supported manifests.

I am going to create a followup pr to include the other OLM types that we need to support.

kevinrizza:
Closing this pr in favor of https://github.com/operator-framework/operator-courier/pull/141 due to an issue with the CI configuration in Travis

PR #140, opened 2019-06-21
Ignore extra yaml files in manifest dir

kevinrizza:
* Update the identifier to return special unknown type when parsing
unexpected yaml file
* Update callers to ignore file when filetype returned is unknown
* Update tests to reflect this change


kevinrizza:
closing due to CI issue

PR #151, opened 2019-08-08
Filter out a black list of files from manifest bundle

galletti94:
/cc @kevinrizza @shawn-hurley

### DIFF
--- a/operatorcourier/manifest_parser.py
+++ b/operatorcourier/manifest_parser.py
@@ -105,3 +105,15 @@ def get_csvs_pkg_info_from_root(source_dir: str) -> Tuple[List[Tuple], Tuple]:
 def is_yaml_file(file_path):
     yaml_ext = ['.yaml', '.yml']
     return os.path.splitext(file_path)[1] in yaml_ext
+
+
+# removes files_names from folder_path and all subdirectories
+def filterOutFiles(folder_path, file_names):
+    for item in os.listdir(folder_path):
+        item_path = os.path.join(folder_path, item)
+        if not os.path.isfile(item_path):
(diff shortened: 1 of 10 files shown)


operatorcourier/push.py
```diff
@@ -8,9 +8,10 @@
     OpCourierQuayCommunicationError,
     OpCourierQuayErrorResponse
 )
+from operatorcourier.manifest_parser import filterOutFiles
 
 logger = logging.getLogger(__name__)
-
+BLACK_LIST = ["art.yaml", "image-references"]
```
kevinrizza: Can we comment this to describe what it is for?

operatorcourier/push.py
```diff
@@ -28,6 +29,7 @@ def push(self, bundle_dir, namespace, repository, release, auth_token):
         :param auth_token: Authentication token used to push to Quay.io.
         """
         logger.info('Generating 64 bit bundle and pushing to app registry.')
+        filterOutFiles(bundle_dir, BLACK_LIST)
```
galletti94: Note that this removes files from a temporary directory created at runtime (not the actual directory)

operatorcourier/manifest_parser.py
```diff
@@ -105,3 +105,13 @@ def get_csvs_pkg_info_from_root(source_dir: str) -> Tuple[List[Tuple], Tuple]:
 def is_yaml_file(file_path):
     yaml_ext = ['.yaml', '.yml']
     return os.path.splitext(file_path)[1] in yaml_ext
+
+def filterOutFiles(folder_path, file_names):
```
kevinrizza: Can we document this function?

PR #161, opened 2019-09-03
test: OMPS integration tests

estroz:
**Description of changes:**
* tox.ini: add integration env for push/omps tests
* omps-integration.Dockerfile: defines a container to run omps integration tests
* tests/: code and config for omps integration tests
* run_integration_tests.sh: runs integration tests. Useful for prow CI

**Motivation for changes:** operator-courier is an [integral part](https://github.com/release-engineering/operators-manifests-push-service/blob/v8.2/omps/quay.py#L279-L282) of manifest verification and pushing to quay.io in OMPS. Every PR can potentially break OMPS, so the desired changes to operator-courier in a PR should be integrated into and tested with the latest OMPS.

Travis CI tests will continue to function as before. These changes are intended to be used in OpenShift CI.

/cc @gallettilance @shawn-hurley @kevinrizza

### DIFF
--- a/omps-integration.Dockerfile
+++ b/omps-integration.Dockerfile
@@ -1,0 +1,17 @@
+ARG OMPS_VERSION=latest
+FROM quay.io/operator-manifests/omps:${OMPS_VERSION}
+
+# Install operator-courier.
+USER 0
+RUN pip3 install koji
+COPY . /repo
+WORKDIR /repo
+RUN pip3 install -U .
+RUN rm -rf /repo
+
+# Re-install omps with this version of operator-courier.
+WORKDIR /src
+RUN pip3 install -U --no-deps .
+
+USER 1001
+ENTRYPOINT docker/install-ca.sh && gunicorn-3 --workers ${WORKERS_NUM} --timeout ${WORKER_TIMEOUT} --bind 0.0.0.0:8080 --access-logfile=- --enable-stdio-inheritance omps.app:app
--- a/omps-integration.Dockerfile
+++ b/omps-integration.Dockerfile
@@ -3,15 +3,14 @@ FROM quay.io/operator-manifests/omps:${OMPS_VERSION}
 
 # Install operator-courier.
 USER 0
-RUN pip3 install koji
-COPY . /repo
-WORKDIR /repo
+RUN pip3 install koji tox
+COPY . /operator-courier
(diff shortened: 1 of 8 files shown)


omps-integration.Dockerfile
```diff
@@ -0,0 +1,17 @@
+ARG OMPS_VERSION=latest
+FROM quay.io/operator-manifests/omps:${OMPS_VERSION}
```
estroz: A prow CI job in the OMPS repo should run to build and push this image. Should this build always be from OMPS `master`?

omps-integration.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+# Install operator-courier.
+USER 0
+RUN pip3 install koji
+COPY . /repo
+WORKDIR /repo
+RUN pip3 install -U .
+RUN rm -rf /repo
+
+# Re-install omps with this version of operator-courier.
```
estroz: Note that the steps below are dependent on how OMPS currently sets up its image filesystem and how it is intended to be run using `gunicorn`. This might break if OMPS changes either of these things.

csomh:
@estroz small note: OMPS has set of [integration test cases](https://github.com/release-engineering/operators-manifests-push-service/tree/master/tests/integration) defined, and we are running them against a qa instance internally, whenever there is a new release to operator-courier. (They are not part of a CI, b/c we didn't have the time to set that up.)

Would it make sense to use (a subset) of those in OpenShift CI, instead of defining a new (to be maintained) set of test cases?

estroz:
@csomh if it is possible to trigger an OMPS integration test on each operator-courier PR and report the results in that PR, then we should set a prow CI config up for that instead. AFAIK that won't be possible because prow CI is configured to report on a per-repo basis.

@kevinrizza no, but if ^ is possible then I can create one.

kevinrizza:
@estroz I'm still not sure I understand how this is supposed to work then. This needs some configuration to run. Where is that configuration defined? Where do these tests actually run? Travis? Did we set those environment variables there?

estroz:
@kevinrizza these tests will run in prow CI. I will submit a config to [openshift/ci-operator](https://github.com/openshift/release/tree/master/ci-operator) once the tests are merged here.

estroz:
Closing in favor of #164 and follow-up.

PR #165, opened 2019-09-23
Support crd.versions field

kevinrizza:
Update the verify command to respect the latest crd spec with regards to
the deprecation of the version field and replacement with spec.versions


awgreene:
This will LGTM once @gallettilance's comments are addressed.

kevinrizza:
> I think we may need to change the ui validate stuff too ([here](https://github.com/operator-framework/operator-courier/blob/09dd121c54ab49560a909db4b9902f96739a7dbf/operatorcourier/validate.py#L670)) and in the const.py file where `version` is required - this could be breaking if the field isn't there.

@gallettilance All of that stuff is about the spec.version of the CSV, not the CRD. I believe I have updated all the references to the CRD spec.version.

PR #171, opened 2019-10-22
support private quay instance

### LINKED ISSUE #170: Support for private quay instance
For now, this tool works only with quay.io instance.
Red Hat allows deploying private[1] quay instance.
Add support for private quay instance.

[1] - https://access.redhat.com/documentation/en-us/red_hat_quay/3/html-single/deploy_red_hat_quay_-_basic/index

jkandasa:
fixes #170

### DIFF
--- a/operatorcourier/push.py
+++ b/operatorcourier/push.py
@@ -21,7 +21,8 @@ class PushCmd():
     def __init__(self):
         pass
 
-    def push(self, bundle_dir, namespace, repository, release, auth_token):
+    def push(self, bundle_dir, namespace, repository, release, auth_token,
+             quay_host, verify_host):
         """Push takes a bundle and pushes it to the specified app registry repository.
 
         :param bundle_dir: Path to generated local directory that contains the bundle.
@@ -29,6 +30,9 @@ class PushCmd():
         :param repository: Repository name of the application described by the bundle.
         :param release: Release version of the bundle.
         :param auth_token: Authentication token used to push to Quay.io.
+        :param quay_host: Can be 'quay.io' or your private instance hostname.
+        :param verify_host: ``quay_host`` TLS/CA verification.
+                            Either a boolean or a string.
         """
         logger.info('Generating 64 bit bundle and pushing to app registry.')
         filterOutFiles(bundle_dir, BLACK_LIST)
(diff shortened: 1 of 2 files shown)


operatorcourier/push.py
```diff
@@ -21,14 +21,18 @@ class PushCmd():
...
         """Push takes a bundle and pushes it to the specified app registry repository.
 
         :param bundle_dir: Path to generated local directory that contains the bundle.
         :param namespace: Namespace that contains the repository for the application.
         :param repository: Repository name of the application described by the bundle.
         :param release: Release version of the bundle.
         :param auth_token: Authentication token used to push to Quay.io.
+        :param quay_host: Can be 'quay.io' or your private instance hostname.
+        :param verify_host: ``quay_host`` TLS/CA verification.
+                            Either a boolean or a string.
```
MartinBasti: I'd put here a full documentation, or refer to requests documentation what usage of `string` means here.
From requests 
```
    :param verify: (optional) Either a boolean, in which case it controls whether we verify
            the server's TLS certificate, or a string, in which case it must be a path
            to a CA bundle to use. Defaults to ``True``.
```

jkandasa: @MartinBasti Thanks! I have updated the documentation.

operatorcourier/push.py
```diff
@@ -21,14 +21,20 @@ class PushCmd():
...
         :param release: Release version of the bundle.
         :param auth_token: Authentication token used to push to Quay.io.
+        :param quay_host: Can be 'quay.io' or your private instance hostname.
+        :param verify_host: Either a boolean, in which case it controls whether we verify
+                            the server's TLS certificate, or a string, in which case
+                            it must be a path to a CA bundle to use.
+                            Defaults to ``True``.
         """
         logger.info('Generating 64 bit bundle and pushing to app registry.')
         filterOutFiles(bundle_dir, BLACK_LIST)
```
kevinrizza: Just below this, don't you also need to update the call to `_push_to_registry()` to actually pass the new parameters along?

operatorcourier/push.py
```diff
@@ -45,14 +51,16 @@ def _create_base64_bundle(self, bundle_dir, repository):
             result64 = base64.b64encode(result).decode("utf-8")
             return result64
 
-    def _push_to_registry(self, namespace, repository, release, bundle, auth_token):
-        push_uri = 'https://quay.io/cnr/api/v1/packages/%s/%s' % (namespace, repository)
+    def _push_to_registry(self, namespace, repository, release, bundle, auth_token,
+                          quay_host, verify_host):
+        push_uri = 'https://%s/cnr/api/v1/packages/%s/%s' % (quay_host, namespace,
```
kevinrizza: If we are going to do this, rather than just support additional `quay.io/cnr` endpoints, why not just make the entire root path of the host to the API configurable? i.e.:

```
push_uri = 'https://%s/api/v1/packages/%s/%s' % (host, namespace,repository)
```

`https://quay.io/cnr` is the full arbitrary root path, the `cnr` is not part of the app-registry spec. Here's a full client implementation for reference: https://github.com/operator-framework/go-appr

PR #177, opened 2019-12-11
Validate version must be the first item of versions

clyang82:
Signed-off-by: Chun Lin Yang <clyang@cn.ibm.com>
>
apiextensions.k8s.io/v1beta1, there was a `version` field instead of `versions`. The `version` field is deprecated and optional, but if it is not empty, it must match the first item in the `versions` field.

This PR is to validate this case.

### DIFF
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -163,12 +163,16 @@ class ValidateCmd():
                 if "versions" in crd['spec']:
                     if "version" in crd['spec']:
                         if "name" in crd['spec']['versions'][0]:
-                            if crd['spec']['version'] != crd['spec']['versions'][0]['name']:
-                                self._log_error("crd spec.version must be the first element in crd spec.versions")
+                            if crd['spec']['version'] != \
+                                    crd['spec']['versions'][0]['name']:
+                                self._log_error("crd spec.version must be "
+                                                "the first element in crd spec.versions")
                                 valid = False
                         else:
-                            if crd['spec']['version'] != crd['spec']['versions'][0]:
-                                self._log_error("crd spec.version must be the first element in crd spec.versions")
+                            if crd['spec']['version'] != \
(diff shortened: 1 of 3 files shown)


kevinrizza:
@clyang82  This pr seems great! Thanks for the contribution. However, fyi we are trying to avoid updating and releasing this project if at all possible given the operator-framework's efforts to deprecate and replace all of our use of app-registry -- so while this pr may merge we may not cut a release for this anytime soon. Is this something you are looking to actively validate?

If you're just looking to contribute to a validation project in the operator-framework, this project aims to replace most of the `operator-courier verify` functionality once app-registry is deprecated: https://github.com/operator-framework/api/tree/master/pkg/validation

/lgtm
/approve

clyang82:
@kevinrizza Thanks for your information. That is OK. I do not have urgent business requirement to do validation. I just found that the yaml file did not work after pass the `courier` validation. If you think the PR is ok to merge, then  just merge it. Thanks.

I will take a look at https://github.com/operator-framework/api/tree/master/pkg/validation

BTW, where I can find more information about app-registry is deprecated? Thanks

kevinrizza:
@clyang82 Please take a look at these enhancements for more details:

https://github.com/openshift/enhancements/blob/master/enhancements/olm/operator-registry.md
https://github.com/openshift/enhancements/blob/master/enhancements/olm/operator-bundle.md

Essentially we are proposing the use of container image registries as storage to host operator bundles and registry index images.

euld1984:
The statement "Validate version must be the first item of versions" [suggests](https://www.youtube.com/@celebritiesageinfo) that when listing out different versions of something, the first item in the list should always be the validated version. This may be important in certain contexts where the order of versions matters, such as in software development where version control is crucial. By placing the validated version first in the list, it helps ensure that it is the most current and reliable version being used.

Issue #188, opened 2020-07-13
[regression?] operator-courier 2.1.8 falsely(?) fails --ui_validate_io check


leifmadsen:
My CI started failing today on the `operator-courier verify --ui_validate_io` check for my Ansible Operators (probably not related to Ansible).

Version 2.1.7 does not result in a failure. When running 2.1.8 I get the following error on my checks:

```
operator-courier verify --ui_validate_io deploy/olm-catalog/service-telemetry-operator/
ERROR: CSV.spec.crd.owned.version isnot in CRD.spec.versions list [0.1.1/service-telemetry-operator.v0.1.1.clusterserviceversion.yaml]
ERROR: CSV.spec.crd.owned.version isnot in CRD.spec.versions list [1.0.0/service-telemetry-operator.v1.0.0.clusterserviceversion.yaml]
ERROR: CSV.spec.crd.owned.version isnot in CRD.spec.versions list [1.0.1/service-telemetry-operator.v1.0.1.clusterserviceversion.yaml]
ERROR: CSV.spec.crd.owned.version isnot in CRD.spec.versions list [1.0.2/service-telemetry-operator.v1.0.2.clusterserviceversion.yaml]
ERROR: CSV.spec.crd.owned.version isnot in CRD.spec.versions list [1.0.3/service-telemetry-operator.v1.0.3.clusterserviceversion.yaml]
Resulting bundle is invalid, input yaml is improperly defined.
```

CSVs are available here:  https://github.com/infrawatch/service-telemetry-operator/tree/master/deploy/olm-catalog/service-telemetry-operator/1.0.2 (version 1.0.2 as an example).

Looking at what the error is telling me, it seems like everything is fine?

CSV --> spec --> CRD --> owned --> version == https://github.com/infrawatch/service-telemetry-operator/blob/master/deploy/olm-catalog/service-telemetry-operator/1.0.2/service-telemetry-operator.v1.0.2.clusterserviceversion.yaml#L93

CRD --> spec --> versions == https://github.com/infrawatch/service-telemetry-operator/blob/master/deploy/olm-catalog/service-telemetry-operator/1.0.2/infra.watch_servicetelemetrys_crd.yaml#L16-L19

jmazzitelli:
I'm seeing the same - I haven't changed the community metadata for my operator in a while - and just today when an unrelated PR got submitted and travis ran the CI tests, I get this:

```
[...]

PR #189, opened 2020-07-13
Correctly validate crd.spec.versions

tiraboschi:
Correctly validate crd.spec.versions

crd.spec.version got deprecated for crd.spec.versions
but in this case we are going to access a list of
dictionaries where the relevant value
is in the name field.

Fix all the validations on spec.versions and add
additional validations.

Fix also its test.

Signed-off-by: Simone Tiraboschi stirabos@redhat.com

Fixes: #188

### DIFF
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -156,11 +156,52 @@ class ValidateCmd():
                 if "group" not in crd['spec']:
                     self._log_error("crd spec.group not defined.")
                     valid = False
-                if "versions" not in crd['spec']:
-                    if "version" not in crd['spec']:
-                        self._log_error("crd spec.version or spec.versions not defined")
+                if (
+                    "versions" not in crd['spec'] and
+                    "version" not in crd['spec']
+                ):
+                    self._log_error(
+                        "crd spec.version or spec.versions not defined."
+                    )
+                    valid = False
+                if "versions" in crd['spec']:
+                    if not len(crd['spec']['versions']) > 0:
+                        self._log_error("crd spec.versions is empty.")
                         valid = False
-
+                    else:
+                        for ver in crd['spec']['versions']:
+                            if (
+                                "name" not in ver or
+                                "served" not in ver or
+                                "storage" not in ver
+                            ):
+                                self._log_error(
+                                    "crd spec.versions contains an invalid "
+                                    "CustomResourceDefinitionVersion."
+                                )
+                                valid = False
+                        if "version" in crd['spec']:
+                            if (
+                                "name" in crd['spec']['version'][0] and
+                                crd['spec']['version'][0]['name'] !=
+                                crd['spec']['version']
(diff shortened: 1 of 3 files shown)


operatorcourier/validate.py
```diff
@@ -273,7 +273,10 @@ def _csv_spec_validation(self, spec, bundleData):
                     if 'version' in csvOwnedCrd:
                         if 'spec' in crd:
                             if 'versions' in crd['spec']:
-                                if csvOwnedCrd['version'] not in crd['spec']['versions']:
+                                if csvOwnedCrd['version'] not in [
+                                    v['name'] for v in crd['spec']['versions']
+                                    if 'name' in v
+                                ]:
                                     self._log_error('CSV.spec.crd.owned.version is'
```
jmazzitelli, line 280: Just a trivial fix to a minor thing - when this error spits out - there should be a space between "is" and "not" but instead it prints out as "... version isnot in CRD..."

```suggestion
                                    self._log_error('CSV.spec.crd.owned.version is '
```

tiraboschi: Done, thanks.

huzefa51:
Hello,
I am trying to push the manifest but getting below error.

ERROR: CSV.spec.crd.owned.version isnot in CRD.spec.versions list [manifests/myapp-operator.clusterserviceversion.yaml]

# operator-courier --version
2.1.8

operatorcourier/validate.py
```diff
@@ -273,8 +273,11 @@ def _csv_spec_validation(self, spec, bundleData):
                     if 'version' in csvOwnedCrd:
                         if 'spec' in crd:
                             if 'versions' in crd['spec']:
-                                if csvOwnedCrd['version'] not in crd['spec']['versions']:
-                                    self._log_error('CSV.spec.crd.owned.version is'
+                                if csvOwnedCrd['version'] not in [
```
waynesun09, line 276: This should also applied to line 284

tiraboschi: I don't think so:
crd['spec']['versions'] is a list of dict where the relevant value is defined by the name key, crd['spec']['version'] is instead a plain text value.

waynesun09: Just checked again, you are right

tiraboschi: but you are right, there is another issue

tiraboschi:
I updated https://github.com/tiraboschi/operator-courier/tree/fix_188 but strangely I still my latest changes here

jmazzitelli:
> I updated https://github.com/tiraboschi/operator-courier/tree/fix_188 but strangely I still my latest changes here

Github is having problems - I saw similar odd behavior in unrelated PRs - just have to wait for them to fix the infrastructure: https://www.githubstatus.com/

jhrozek:
btw the patch works for me, thank you for the work!

PR #194, opened 2020-08-18
courier: fix validation with multiple CRD versions

fromanirh:
If we ship a bundle which have multiple CRD version, without this patch
the validation may fail with a false negative with

  CRD.spec.version does not match CSV.spec.crd.owned.version

This is because we check that EVERY owned CRD matches with the
supported version in the CSV.
But this doesn't work if we supply a CSV which points to a CRD version
(example: v1) which *is* in the bundle, but which also contains
older CRD versions.

IOW the failed scenario is:
```
CSV:
  owns: v1, v1alpha1
CRD:
  provides: v1, v1alpha1
  supports: v1
```
here "supports" mean the value listed in the `version` field of the CRD.

In this case the need to check:
1. the CRD version (v1) is owned by the CSV
2. all the versions owned in the CSV are declared in the CRD

The check #2 was already guarantee by the code looping over CSV owned
CRD versions; this patch fixes the code to fix the check #1.

Related-To: https://bugzilla.redhat.com/show_bug.cgi?id=1869525
Signed-off-by: Francesco Romani <fromani@redhat.com>

PR #195, opened 2020-08-19
courier: fix validation with multiple CRD versions

fromanirh:
If we ship a bundle which have multiple CRD version, without this patch
the validation may fail with a false negative with

  CRD.spec.version does not match CSV.spec.crd.owned.version

This is because we check that EVERY owned CRD matches with the
supported version in the CSV.
But this doesn't work if we supply a CSV which points to a CRD version
(example: v1) which *is* in the bundle, but which also contains
older CRD versions.

IOW the failed scenario is:
CSV:
  owns: v1, v1alpha1
CRD:
  provides: v1, v1alpha1
  supports: v1

In this case the need to check:
1. the CRD version (v1) is owned by the CSV
2. all the versions owned in the CSV are declared in the CRD

The check #2 was already guarantee by the code looping over CSV owned
CRD versions; this patch fixes the code to fix the check #1.

Signed-off-by: Francesco Romani <fromani@redhat.com>

### DIFF
--- a/operatorcourier/validate.py
+++ b/operatorcourier/validate.py
@@ -1,3 +1,4 @@
+import collections
 import logging
 import json
 import semver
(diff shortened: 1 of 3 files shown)


fromanirh:
ok, the intent was triggering the stale tests, it seems that closing/reopening in #194 did the trick, so closing this clone for now.

